PR #10, opened 2020-02-24
Bootleg global subsystems vars improvement.

Ghommie:
Ripped from the StonedMC made by MrStonedOne.

Should I port the real thing later or stay true to the built from scratch as much as possible?
Also some of these subsystems should be made into global lists later.


code/_core/mob/living/advanced/human/npc/syndicate.dm
```diff
@@ -76,9 +76,8 @@
 
 	add_outfit(outfit_to_use)
 
-	var/picked_weapon = pick(possible_weapons)
-
 	/*
+	var/picked_weapon = pick(possible_weapons)
```
Ghommie: Unrelated, but this bit was throwing a warning because it was unused. Annoying.

BurgerLUA: oh yeah

PR #14, opened 2020-02-26
Turned typecheck procs into macros.

Ghommie:
Title. A better choice I would say, also because it's very unlikely for typechecks to be ever used for `call()` anyway.

Plus some required 512-513 compatibility since I dropped in some `islist()` (the one introduced in 513, without the underscore) calls while PR associative list improvements.


BurgerLUA:
absolute madlad with the work, keep it up

PR #29, opened 2020-06-26
Another big small slavic update

Ecologistos:
(Hope rev antags will work, because I couldn't test them myself.)
uuuuuh
I think I added something, but I'm in a shitty condition rn and can't really remember what.
Some AK-13 change, some sprites and new armor, new antag that might or might not work, also unique sounds for weapons and a gauss rifle sound for later gauss addition.
Stuff compiles tho.


Ecologistos:
That shouldn't happen

Ecologistos:
A lot of stuff got changed after this PR, so I'd rarther close it and change the current version.
Also might edit and balance stuff a little more.

PR #30, opened 2020-06-28
Added a 7.62 Rifle

PigeonLord:
Added a 7.62 Semi-Auto Rifle with low heat and heat buildup. The purpose of the Rifle is to have a stepping stone between the Sniper and the LMG which performs well at all ranges but lacks the capability to lay automatic fire like the LMG or the Added Zoom of the Sniper Rifle.


PigeonLord:
(Note that I haven't added it to Shops or the .dme because A)I'm not sure that this Rifle will be accepted and B)I don't really know how to interact with those systems, as in other servers i've worked with I didn't have to add new items to the DME manually, nor I have much experience in mapping.)

PigeonLord:
This PR conflicts severely with new PRs, so i'm deleting it and restarting from scratch.

PR #33, opened 2020-06-30
Still worse at penetrating, but 25 Pierce was useless against any kind of armor, especially for 7.62 which dealt more Pierce than Blunt.

PigeonLord:
Raised the .308 and Short 7.62 to 40, and the 7.62 to 55 to account for better penetration. Thought the smaller falloff was a nerf enough for the Short 7.62, will change on request.

### DIFF
--- a/code/_core/datum/damagetype/ranged/bullet/rifle_308.dm
+++ b/code/_core/datum/damagetype/ranged/bullet/rifle_308.dm
@@ -10,7 +10,7 @@
 	//How much armor to penetrate. It basically removes the percentage of the armor using these values.
 	attack_damage_penetration = list(
 		BLUNT = 50,
-		PIERCE = 25
+		PIERCE = 40
 	)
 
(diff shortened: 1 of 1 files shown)


PigeonLord:
For the record, against a 75/75 armor, 7.62 deals literally 0 damage before this change.

BurgerLUA:
I actually think it's pretty balanced right now.
image

7.62mm is meant to do a lot of damage against unarmored foes as it's a large projectile, while still really out-damaging the 5.56 against standard armor, while 5.56mm is meant to pierce heavily armored foes like those in hardsuits. I don't think I will be merging this PR unless there is is a decent amount of support for this.

PigeonLord:
Looking at the numbers better, you have a fair point, but I think that the 7.62 should have it's Blunt and Pierce Pen swapped - 25 Brute, 50 Pierce, since it deals more pierce.

PR #36, opened 2020-06-30
Added the Medical Voidsuit and the SEVA suit (Please merge)

PigeonLord:
I had to put them both in one PR since all the Hardsuit helmets are in one file.


PigeonLord:
Split the Monolith and the Regular variant. The person who requested the SEVA suit said they'd want the regular variant to be available for sale, and the Monolith Variant to be found as loot

PR #45, opened 2020-07-02
Makes the NATO rounds tooltip actually true, Fixes the bulletproof armor

PigeonLord:
Change List:
1)The Bulletproof armor is only used by mannequins, and it having different values for Blunt and Pierce pen gives you a very poor idea of how strong a weapon actually is.
2)NATO rounds are now all Pierce damage, and weaker than Cartridge-designated rounds by 10 damage. This makes them worse against unarmored targets, but better against armored ones.
3)556 Retains better damage and DPS against Heavy armor (in the order of 75 pen), while 762 excels against unarmored or lightly armored targets.

### DIFF
--- a/code/_core/datum/damagetype/ranged/bullet/rifle_223.dm
+++ b/code/_core/datum/damagetype/ranged/bullet/rifle_223.dm
@@ -20,14 +20,12 @@
 
 	//The base attack damage of the weapon. It's a flat value, unaffected by any skills or attributes.
 	attack_damage_base = list(
-		BLUNT = 10,
 		PIERCE = 30
 	)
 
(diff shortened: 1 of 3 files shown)


PigeonLord:
I also suggest putting in additional mannequins with heavy armor (75 brute 75 pierce) to further give them an idea on what weapons they should use against what targets.

PigeonLord:
There's a tooltip that says that NATO rounds do less damage but pierce armor more. I thought the easiest way to make that actually true was to switch NATO to pure pierce, to avoid the headache of balancing the two damage types.
If that's a problem, I could fuck with the numbers more to find a better solution. Or if you don't intend to actually differentiate ammo types, I could just revert the damage update and keep the Bulletproof Armor thing.

PR #47, opened 2020-07-02
Added the KS23 Shotgun

PigeonLord:


code/_core/obj/item/bullet/shotgun_23.dm
```diff
@@ -0,0 +1,64 @@
+/obj/item/bullet_cartridge/shotgun_23/
```
BurgerLUA, line 1: This is missing a projectile count.

PR #53, opened 2020-07-04
Jungle Hellscape Map

Real-MAGNUM:
Adds a new map, and a shuttle to reach it with along with slight bugfixing to polish the map a bit. only unnecessary addition are the new buckshot alt drum magazines for the auto shotgun. The defaults were slugs, and I found it strange, and a hassle to manually load instead of using the ammo refiller. Skeletons have a new loadout group as miners as well. 

High res map: https://imgur.com/a/tt5z07i

To see the map in full res, magnify it on imgur before saving the image.


BurgerLUA:
Please do not do any map changes. Also, please remove the changes to the area files as an overhaul was made.

PR #68, opened 2020-07-10
Item Mapping

Real-MAGNUM:
Creates shop locations for Sorcerer's robes/cape, and tome of summon skeleton


BurgerLUA:
This PR isn't needed. Map edits are handled by me for now.

PR #76, opened 2020-08-11
New sound for Tungsten Carbine.

Alcubierre7985:
Modified from:
https://freesound.org/people/adcbicycle/sounds/13854/

Licensed under Creative Commons 0 (CC0)


BurgerLUA:
I don't know if this would be a good fit for the tungsten carbine as it's more like a railgun, but I'm sure this can be good for other weapons.

PR #77, opened 2020-09-03
Energy Sword Recoloring

pitaden:
Man, I really gotta stop doing these PRs on my master branch

By clicking on energy weapons with a multitool, you can color their parts like you would with a custom dye core and clothing.


pitaden:
keeping this as a draft until i make sure this won't mess with people's saves (specifically, energy shields)

pitaden:
the conflict now seems to just be a comment and a line of `polymorphs` where there's nothing in the current master, so it shouldn't be a problem

BurgerLUA (approved):

PR #79, opened 2020-09-20
another cool pr

AdrienTheMan:
-puts a bandaid on a booboo so energy swords can now save their color from using a multitool
-added a portable barbecue grill and possibly spaghettified code (can only cook by putting things inside due to my limited copy pasting skills not being able to prevent in-motion grillin action issues)
-fixed a singular typo on some pants (:
yea
feat. cozy grill lighting


BurgerLUA (approved):
make sure to atomize PRs next time

PR #81, opened 2020-09-20
fix

Ecologistos:
He is stupid
Fixes offset and AK-12 sprite.


Stalkeros2:
Also makes it less of an LMG, restricting the firerate to ~750 RPM, which is somewhat close to the actual one.

BurgerLUA (approved):

PR #83, opened 2020-09-27
this took me five fucking hours

AdrienTheMan:
-fixes being able to equip paper bags containing items
~~-gives a little feature where you can switch through like 9 cool logos when alt clicking!!!!!!!! (however i couldnt get fancy polymorphic magic so the bag & logo get painted together if you dye it)~~
-paper bags are now polymorph
-they can receive a logo or pattern by alt clicking the air with help and disarm intent
-yes the logos save
-2 extra presets, NT and syndie with their respective colors
yas
image outdated but there's a newer one lower
**these additions are necessities in these trying times**


code/_core/obj/item/clothing/hats/paperbag.dm
```diff
@@ -20,22 +20,80 @@
...
+
+/obj/item/clothing/head/hat/paperbag/save_item_data(var/save_inventory = TRUE)
+	. = ..()
+	SAVEVAR("logo")
+	return .
+
+/obj/item/clothing/head/hat/paperbag/load_item_data_pre(var/mob/living/advanced/player/P,var/list/object_data)
+	. = ..()
+	LOADVAR("logo")
+	update_sprite()
```
BurgerLUA, line 33: update_sprite should be called in `Finalize()` instead.

BurgerLUA (changes requested):
Looks good, just needs one change.

code/_core/obj/item/clothing/hats/paperbag.dm
AdrienTheMan: is this good now i dont have a big brain

AdrienTheMan: okay nevermind i dont even need to call anything to update or finalize the sprite, it loads it up properly without

AdrienTheMan:
okay so the bags are now polymorph compatible, can cycle through 5 logos with help intent alt click and 1 pattern (the stripe) with disarm alt click

BurgerLUA (approved):
Not what I meant by the requested changes but I'll update it myself.

PR #89, opened 2020-10-25
minecraft steve armour

Davidththirdwhat:
credits to mojang for minecraft and credits to the unknown artist who made the sprites because I cant find whoever the hell made these

pretty much just mk-1 syndi hardsuit but with no slowdown


BurgerLUA (approved):
Don't know about giving it syndie armor resistances but an armor rework is coming anyways so

PR #95, opened 2020-11-03
MREs and lots of food.

Stalkeros2:
(It also adds them on most of the maps, so beware or you might break something idk.)
Implements MREs for each of the main factions, with varying contents.
Credits to Prishtina for gauze's sprite I've used as Syndicate's MRE, Stalkerbuild for food sprites, Burger for edit of the NT's MRE's sprite and The Roma#2717 for giving the original sprites.
Sprites used. There're also opened MREs sprites in the game files, but I really dunno how I would implement that:
image
NT's MRE's contents. As you can see, it is uuuuh idk eatable.
image
Syndicate's MRE's contents. Has slightly more space to NT's MRE due to being the defensive one, not offensive.
image
Revs' MRE's contents. Has the most food available because most of the food was stolen from Stalkerbuild, so not giving it to revs would be strange... and, honestly, because muh bias.
image


Stalkeros2:
Wait.

Stalkeros2:
Darn, that might've been a misclick sorry.

code/_core/obj/item/container/food/package/conserved.dm
```diff
@@ -0,0 +1,106 @@
+/obj/item/container/food/package/conserved
+	icon = 'icons/obj/item/consumable/food/conserved.dmi'
+	packaging = null
```
Stalkeros2, line 3: It might need some other way of implementation, and by this I mean something that's not equal to null.
Couldn't think of anything else, so, yes.
:thinking: Maybe a 'nothing' item could work.

Stalkeros2:
f u c k

Stalkeros2 (commented):
What did I do why is there a review aaaaaaaaa

BurgerLUA:
There are map conflicts. This is why you don't make map changes.

Stalkeros2:
Oh.

Stalkeros2:
Yes, I'll take more time and polish all this stuff before releasing, despite it already being quite... well, working, I guess?
Not like I'm in a hurry.

PR #96, opened 2020-11-13
Fixes Xray Laser

Zembaflemba:
Xray Laser now properly penetrates all armor.

### DIFF
--- a/code/_core/datum/damagetype/ranged/laser/rifle.dm
+++ b/code/_core/datum/damagetype/ranged/laser/rifle.dm
@@ -44,7 +44,7 @@
 
 	//How much armor to penetrate. It basically removes the percentage of the armor using these values.
 	attack_damage_penetration = list(
-		LASER = AP_AXE*0.9,
-		HEAT = AP_AXE*0.1,
-		RAD = AP_AXE*0.1
+		LASER = AP_AXE*9999,
+		HEAT = AP_AXE*9999,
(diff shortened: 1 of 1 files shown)


Zembaflemba:
Also I've tested this on a local server and works correctly. No huge damage numbers or anything.

code/_core/datum/damagetype/ranged/laser/rifle.dm
```diff
@@ -44,7 +44,7 @@
 
 	//How much armor to penetrate. It basically removes the percentage of the armor using these values.
 	attack_damage_penetration = list(
-		LASER = AP_AXE*0.9,
-		HEAT = AP_AXE*0.1,
-		RAD = AP_AXE*0.1
-	)
\ No newline at end of file
+		LASER = AP_AXE*9999,
+		HEAT = AP_AXE*9999,
```
BurgerLUA, line 48: Just for conformity, all large values should be written as INFINITY as this specific number doesn't run armor checks instead of running armor checks with large values.

BurgerLUA (changes requested):

code/_core/datum/damagetype/ranged/laser/rifle.dm
Zembaflemba: 👍

BurgerLUA (approved):

PR #97, opened 2020-11-14
sprite adjusment for turret

Davidththirdwhat:
the turret shouldn't look like it is going up and down now when it's turning also small adjustment to the dead sprite


BurgerLUA:
Oh I actually made a new variant of the turret but I can always add another type.

BurgerLUA (approved):

PR #107, opened 2021-01-07
added floor descriptions

chesse20:
for the lore loves who love to examine every tile

# What this PR does
Adds descriptions to many of the turfs in the game

I've been trapped inside my house for 9 months and I need to make commits to github to make myself feel useful, also gives players something to read while running around doing stuff

### DIFF
--- a/code/_core/turf/simulated/floor/astroturf.dm
+++ b/code/_core/turf/simulated/floor/astroturf.dm
@@ -3,5 +3,7 @@
 	name = "astroturf"
 	icon = 'icons/turf/floor/natural.dmi'
 	icon_state = "grass"
+	desc = "Fake grass"
+	var/desc_extended = "it doesn't feel pain when you step on it, unlike real grass"
 
 	footstep_id = "grass"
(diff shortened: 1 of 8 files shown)


code/_core/turf/simulated/floor/astroturf.dm
```diff
@@ -3,5 +3,7 @@
 	name = "astroturf"
 	icon = 'icons/turf/floor/natural.dmi'
 	icon_state = "grass"
+	desc = "Fake grass"
+	var/desc_extended = "it doesn't feel pain when you step on it, unlike real grass"
```
BurgerLUA, line 7: ``var/desc_extended`` caused compile errors because it's already defined. It should be just ``desc_extended``

BurgerLUA (changes requested):

code/_core/turf/simulated/floor/astroturf.dm
chesse20: ah whoops, went over the code real quick and removed those

BurgerLUA (approved):

PR #112, opened 2021-01-28
(It broke(((( ) Rev miniupdate

Stalkeros2:
# What this PR does
Resprites Slav-Safe armor and adds a thermonocle for them, which, apparently, doesn't work. Gives balaclavas/ski masks to "better-than-basic-but-still-basic" rev troops, as well as antag. Expands Slav-Safe's protection zones a little, because new sprite has groin protection.

Current Slav-Safe's sprite is Bay12's SWAT armor's, which isn't that cool.
image


Stalkeros2:
It broke(((((

PR #117, opened 2021-02-06
fishing fixes part 841147328942

AdrienTheMan:
# What this PR does
-ahoy me matey we be patchin arghhhhh
-possibly fixes endless bait dupe (and makes it not stackable)
-bandaids the easy bobber as it was disgustingly bad previously
-adds a spam check to the fishing rod and a short moment before fish can start biting (its a singular second dw)
-fixes the fishing rod sprite not updating when simply removing or losing the line
-gives fishable tiles their actual reward pools as they would all default to sea previously
https://user-images.githubusercontent.com/9487319/107103107-c5182900-681c-11eb-8a18-106e1ee46569.mp4

### DIFF
--- a/code/_core/obj/item/fishing/rod.dm
+++ b/code/_core/obj/item/fishing/rod.dm
@@ -26,6 +26,7 @@
 	var/catch_time
 
 	var/next_tick = 0
+	var/nospam
 
 	value = 200
 
@@ -72,6 +73,7 @@
 		else if(line)
 			object_removed = line
 			line = null
+			update_sprite()
 		if(!object_removed)
 			caller.to_chat(span("notice","There is nothing to remove from \the [src.name]."))
 		caller.to_chat(span("notice","You remove \the [object_removed.name]."))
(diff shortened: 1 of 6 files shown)


code/_core/obj/item/fishing/rod.dm
```diff
@@ -26,6 +26,7 @@
 	var/catch_time
 
 	var/next_tick = 0
+	var/nospam
```
BurgerLUA, line 29: Instead of no_spam, you instead should use
```
INTERACT_CHECK
INTERACT_DELAY
```
before the code executes.

BurgerLUA (changes requested):

code/_core/obj/item/fishing/rod.dm
AdrienTheMan: reasons for that are in the #contributing channel of the discord, do tell if more changes are required tho

BurgerLUA (approved):

PR #127, opened 2021-02-11
nice fishing things aagain

AdrienTheMan:
# What this PR does
-adds a telescopic fishing rod, which you can extend and retract to fit in a tackle box
-adds it to the lava rare pool but feel free to add it anywhere else idk im feelin lazy
-adds a little ! icon when fish bites (: very comfy
epic


code/_core/obj/item/fishing/rod.dm
```diff
@@ -17,6 +17,7 @@
 	var/obj/item/fishing/bait/bait
 
 	var/obj/effect/fishing_bob/fishing_bob
+	var/obj/effect/alert/exclaim/fishing_alert
```
BurgerLUA, line 20: It would be better if you just made this a temp effect so you don't have to store the alert effect here.

code/_core/obj/item/fishing/rod.dm
```diff
@@ -27,6 +28,7 @@
 
 	var/next_tick = 0
 	var/nospam
+	var/compact = FALSE
```
BurgerLUA, line 31: compact should be saved to the savefile. Check out the _magazine.dm file for how items are saved.

BurgerLUA (changes requested):

code/_core/obj/item/fishing/rod.dm
AdrienTheMan: not sure it'd be good as its currently called by multiple functions so it can get deleted as soon as it is reeled in and such as they stay for a couple seconds, does this require a rewrite or no?

code/_core/obj/item/fishing/rod.dm
AdrienTheMan: most people are gonna store it inside tackle boxes which they have to be required to be compact = false for that to begin with, im not sure its important add it to the save file? correct me if im wrong

code/_core/obj/item/fishing/rod.dm
BurgerLUA: I guess your code is fine.

code/_core/obj/item/fishing/rod.dm
BurgerLUA: I think burgerstation has a safety where if an item is too big to fit in something, it drops it when someone spawns. Not sure though but I rather be on the safe side.

AdrienTheMan: done

BurgerLUA (approved):

BurgerLUA:
Will merge this in about 5 hours or so.

BurgerLUA (approved):

PR #130, opened 2021-02-12
paper bag logo menu!!!!!! lets goo

AdrienTheMan:
# What this PR does
-makes changing paper bag logos less confusing with a simple menu opened from an altclick
-rewrites my babycode into toddlercode (: 
-worn paperbags now do apply colors
lunch


BurgerLUA (changes requested):
There should generally be a INTERACT_CHECK (but not an INTERACT_DELAY`) after every input to confirm that the user can still interact with the object. input() halts the code and waits for user input, and some things, such as user position and whether or not that object still exists, can change.

Other than that, this PR looks good.

code/_core/obj/item/clothing/head/helmet/full/paper_bag.dm
```diff
@@ -45,35 +45,48 @@
...
+			caller.to_chat(span("notice","You decide not to change \the [src.name]'s design."))
+			return TRUE
 		update_sprite()
-
-
-	return TRUE
-
-
+		return TRUE
+	else . = ..()
```
BurgerLUA, line 88: Uh this code looks wonky as heck but I guess I can clean it up.

BurgerLUA (approved):

PR #131, opened 2021-02-12
weed time

AdrienTheMan:
# What this PR does
-brings all the weed sprites (along some cigarette ones, all credits to goon iirc)
-just weed. thats it. nothing else, no smoking or whatever. just weed
-weed make you HUNGRY so basically this is a way to make you not fat (but its unhealthy anyway)
-life weed and death weed (: 
we'll have smoking one day probably idk
sidenote poisons do not deal any damage at the moment but who cares **DRUG**
weed lol
what


Autisem:
hm yes death weed burger

Jokes aside might become fun for chemistry/alchemy

code/_core/datum/reagent/reagent_nutrition_plants.dm
```diff
@@ -149,3 +149,15 @@
 	return .
 
 
+/reagent/nutrition/lipolicide
+	name = "lipolicide"
+	desc = "Plant-based nutrition-eating germs."
+	color = "#511500"
+
+	nutrition_amount = -5
+	nutrition_quality_amount = -1
```
BurgerLUA, line 158: Not sure if this will behave well as a negative number. This should be set to 0 as a precaution because I'm pretty sure it would break things here or best case scenario consider lipocide healthy.

BurgerLUA (approved):

PR #141, opened 2021-02-17
Goblin speed increase (for real this time)

Real-MAGNUM:
# What this PR does
doubles the previous movespeed of goblins (quadruples current)

I'm dumb and read move delay as move speed


code/_core/mob/living/simple/goblin.dm
```diff
@@ -25,7 +25,7 @@
 	blood_type = /reagent/blood/goblin
 	blood_volume = 300
 
-	move_delay = AI_TICK*2
```
BurgerLUA, line 28: This should just be AI_TICK as AI_TICK is the fastest the AI can process, so that should generally be its speed.

BurgerLUA (changes requested):

BurgerLUA:
PR has been stale for a bit.

PR #147, opened 2021-02-21
Rev Update (kang doesn't fuck up as hard edition)

KangTut:
# What this PR does
Gives rev antags socks

Revs don't have socks for some reason. This fixes that to give them a minor speed boost.

### DIFF
--- a/code/_core/datum/loadout/rev_new.dm
+++ b/code/_core/datum/loadout/rev_new.dm
@@ -159,6 +159,8 @@
 		/obj/item/clothing/head/hat/skimask/black,
 		/obj/item/clothing/hands/gloves/colored/padded/black,
 		/obj/item/clothing/hands/gloves/colored/padded/black/left,
+		/obj/item/clothing/feet/socks/knee,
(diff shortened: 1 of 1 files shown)


Stalkeros2:
Hell yes.

Moltovz:
I feel like Rev's slow speed is a pretty integral part of their kit, and makes them feel balanced. maybe that's just me?

KangTut:
They'll still be fairly slow compared to syndies who can get a 20% speed boost if they buy the mk3 armor. Two socks will only increase your speed by 10%. Only way a rev is going to match that speed is if one of their ally NPCs with winter boots dies and you loot them. 

I also see revs more as a hit and run antag. Their guns aren't very good and either function well at very short range or long range both of which need you to be quick to do well in.

BurgerLUA (approved):

BurgerLUA:
If you told me that 2 years ago that I'd be running a server and that players were having a serious debate on whether or not revs should have socks or not for balance purposes, I wouldn't believe you.

PR #155, opened 2021-02-22
BoH & SoH buff

speenor:
# What this PR does
BoH new max item size is 15, so you can store heavier items like armor. Much-needed to make it more viable then the Syndie duffelbag. 
SoH new max item size is 10, you can store heavier items, but not the heaviest.

People usually use the Syndie bag over the BoH because that extra line of inventory space isn't too useful compared to the zero-slowdown of the Syndie bag. Plus, a way to store things like powerarmor for selling/storage is nice.


mrpibb4201337:
Wow he actually did code it

speenor:
yeah but it doesn't fucking work so i have to fix it. i'll just make a new PR where it works

PR #156, opened 2021-02-22
BoH and SoH buffs

speenor:
# What this PR does
BoH can now hold size 15 items. This makes it so you can store heavier armors.
SoH can now hold size 12 items. Not as heavy as the BoH but still, you can store some heavy stuff.

BoH and SoH aren't really used, because people prefer the zero-slowdown of the Syndie bags. This PR makes them viable for use in trading/hauling/storage and more shit.

### DIFF
--- a/code/_core/obj/item/clothing/back/storage/bluespace/_bluespace.dm
+++ b/code/_core/obj/item/clothing/back/storage/bluespace/_bluespace.dm
@@ -7,7 +7,7 @@
 
 
 	dynamic_inventory_count = MAX_INVENTORY_X*4
-	container_max_size = SIZE_5
+	container_max_size = SIZE_5*3
 
 	size = MAX_INVENTORY_X*4*SIZE_5
 
@@ -27,7 +27,7 @@
 	icon = 'icons/obj/item/clothing/back/satchel/bluespace.dmi'
 
 	dynamic_inventory_count = MAX_INVENTORY_X*4
-	container_max_size = SIZE_4
+	container_max_size = SIZE_4*3
 
 	size = MAX_INVENTORY_X*4*SIZE_4
 


Moltovz:
You know I actually was not aware that the syndie bags had zero slowdown. Nice change though!

code/_core/obj/item/clothing/back/storage/bluespace/_bluespace.dm
```diff
@@ -7,7 +7,7 @@
 
 
 	dynamic_inventory_count = MAX_INVENTORY_X*4
-	container_max_size = SIZE_5
+	container_max_size = SIZE_5*3
```
BurgerLUA, line 10: You should be using the size defines for this (SIZE_5,SIZE_6,SIZE_7) instead of directly multiplying it.

code/_core/obj/item/clothing/back/storage/bluespace/_bluespace.dm
```diff
@@ -27,7 +27,7 @@
 	icon = 'icons/obj/item/clothing/back/satchel/bluespace.dmi'
 
 	dynamic_inventory_count = MAX_INVENTORY_X*4
-	container_max_size = SIZE_4
+	container_max_size = SIZE_4*3
```
BurgerLUA, line 30: You should be using the size defines for this (SIZE_5,SIZE_6,SIZE_7) instead of directly multiplying it.

BurgerLUA (changes requested):

code/_core/obj/item/clothing/back/storage/bluespace/_bluespace.dm
speenor: i tried this, it threw an error

BurgerLUA: What error did it throw?

speenor: it just wouldn't compile. i don't know exactly

BurgerLUA (approved):

PR #157, opened 2021-02-23
Small map fixes

PrimeGenerations:
# What this PR does

Fixes being able to use dropship from outstide.

Makes Charlie Airlock immortal. Bye bye bubblegum corpse destroyer of airlocks.

Fixes

### DIFF
--- a/maps/core/burgerstation.dmm
+++ b/maps/core/burgerstation.dmm
@@ -15,8 +15,8 @@
 /obj/structure/interactive/door/airlock/shuttle,
 /obj/structure/interactive/scanner/living,
 /obj/structure/interactive/conveyor{
-	icon_state = "conveyor";
-	dir = 1
+	dir = 1;
+	icon_state = "conveyor"
 	},
 /turf/simulated/floor/tile/shuttle,
 /area/transit/cargo/ship)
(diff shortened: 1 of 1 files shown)


skigear:
being able to get on the shuttle's back before it takes off isn't really an issue, other than that it can make getting on it easier

adding this would also get rid of jumping into bluespace, which is a mechanic that also doesn't cause much issue

maps/core/burgerstation.dmm
```diff
@@ -2878,8 +2891,11 @@
 /turf/simulated/floor/tile/grey,
 /area/burgerstation/hall/north)
 "ku" = (
-/turf/simulated/floor/plating,
-/area/transit/dropship/bravo/station)
+/obj/structure/interactive/restocker/ammo{
+	anchored = 1
```
BurgerLUA, line 2895: Wait, why is this anchored?

maps/core/burgerstation.dmm
```diff
@@ -205,16 +205,19 @@
 /turf/simulated/floor/tile/dark,
 /area/burgerstation/armory)
 "aO" = (
-/turf/simulated/floor/plating/shuttle,
-/area/transit/dropship/delta/station)
+/obj/structure/interactive/restocker/ammo{
+	anchored = 1
```
BurgerLUA, line 209: this too

BurgerLUA (changes requested):
What specifically did you do to make the airlock immortal? Also why are the ammo restockers anchored now?

BurgerLUA:
PR is 22 days stale with conflicting files, so I'm closing it.

PR #162, opened 2021-02-28
The Genetics PR

BurgerLUA:
# What this PR does

This PR effectively adds genetics to the game in a form of magic loot. If you've played Path of Exile, you'll notice some similarities as this system was based on item improvement via currency.

### Traits
There are 3 types of traits: Positive, Neutral, and Negative. Traits are given categories as well, so you can't have two traits that both modify movement speed.

### Genetic Rarity
Every player character now has genetic rarity. Everyone starts off with common genetic rarity. Genetic Rarity determines how many positive/neutral  traits you're allowed to have. and what orbs you're allowed to use. The values are:

- Common: Max 2 positive/neutral traits
- Uncommon: Max 3 positive/neutral traits.
- Rare: Max 4 positive/neutral traits.
- Mythical: Max 5 positive/neutral traits.
- Legendary: Max 6 positive/neutral traits.

As for what orbs you're allowed to use, you can only use orbs that have a rarity greater or equal to your current genetic rarity. For example if you have a rare genetic quality, you can use Rare, Mythical, and Legendary orbs, but not Uncommon or Common orbs.

## Orbs

Orbs allow the player to modify their traits. They are acquired via loot. Most have a random rarity assigned to them, which affects their function.

### Genetic Orb of Purging
Removes all traits from you. If used when you have no traits, sets your genetic rarity to common. The rarity of this orb is always broken, and any person of any genetic rarity can use it.

### Genetic Orb of Upgrading
Increases your genetic quality to this level, regardless of your previous genetic quality. Can be used if you have no traits. Note that you can no longer use orbs that have a quality below your new level when used, so be careful.

### Genetic Orb of Exalting
Adds a new random trait to your character. Can removing existing traits if the newly added trait conflicts with a previously existing one.

### Genetic Orb of Splicing
Removes a random trait from your character.

### Genetic Orb of Scrambling
Removes all your traits from your character, and adds new ones to your character. The amount of new traits added to your character is based on your current genetic quality.

Players need further ways to improve their characters.

# Todo:

- [x] Make Traits save.
- [x] Improve the sprite.
- [ ] Possibly add penalties if you have too many positive traits but not enough negative ones.
- [ ] Add more traits.
- [x] Add a device that scans the target's traits.
- [ ] Add genetic templates that copy your traits into an orb so players can sell them to other players.


Stalkeros2:
I didn't play Path of Exile tho. :disappointed:

PR #173, opened 2021-03-05
dye dye dye dye dye dye dye dye dye dye dye dye dye

AdrienTheMan:
# What this PR does
-fixes the "Current" color preview of the custom slime core (also fixing where the color choice is on opening it)
-adds the ability to COPY the color of an item onto the custom core with alt clicking for easier coloring
-adds a description to da core
-makes the satchel/backpack/duffles dyable (:
coolphoto

### DIFF
--- a/code/_core/obj/item/dying.dm
+++ b/code/_core/obj/item/dying.dm
@@ -15,14 +15,29 @@
 		return FALSE
 
 	var/choice
-
+	var/mob/E = caller
 	if(length(polymorphs))
+		if(E.attack_flags & CONTROL_MOD_ALT && D.type == /obj/item/slime_core/custom)
+			choice = input("What do you want to copy?","Dye Selection") as null|anything in polymorphs
+			if(choice)
+				INTERACT_CHECK_NO_DELAY(src)
+				INTERACT_CHECK_NO_DELAY(D)
+				D.color = polymorphs[choice]
+				caller.to_chat(span("notice","You copy \the color from \the [src.name]."))
+				return TRUE
+			else return FALSE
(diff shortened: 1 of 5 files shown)


code/_core/obj/item/dying.dm
```diff
@@ -15,14 +15,29 @@
...
+				caller.to_chat(span("notice","You copy \the color from \the [src.name]."))
+				return TRUE
+			else return FALSE
 		choice = input("What do you want to dye?","Dye Selection") as null|anything in polymorphs
 		if(choice)
 			INTERACT_CHECK_NO_DELAY(src)
 			INTERACT_CHECK_NO_DELAY(D)
 			polymorphs[choice] = blend_colors(polymorphs[choice] ? polymorphs[choice] : "#FFFFFF",dye_color,dye_strength)
 	else
+		if(E.attack_flags & CONTROL_MOD_ALT && D.type == /obj/item/slime_core/custom)
```
BurgerLUA, line 35: You should probably add the copy+paste feature to the slime core code instead of here.

code/_core/obj/item/dying.dm
```diff
@@ -15,14 +15,29 @@
 		return FALSE
 
 	var/choice
-
+	var/mob/E = caller
 	if(length(polymorphs))
+		if(E.attack_flags & CONTROL_MOD_ALT && D.type == /obj/item/slime_core/custom)
```
BurgerLUA, line 20: You should probably add the copy+paste feature to the slime core code instead of here.

BurgerLUA (changes requested):

BurgerLUA (approved):

PR #174, opened 2021-03-06
fishing LUCK time

AdrienTheMan:
# What this PR does
-fixes the value of fishing rods to take accesories in count
-luck now impacts fishing!
-fixed telescopic rod "extend now!!!" shenanigans on clicking literally anything
 ~~there is a chance on catch to get a buff giving better loot, affected from attribute and item luck
the chances for the trigger are MINIMUM 1% and MAXIMUM 75%, however item luck is half as good as attribute luck, so basically each points of luck above 50 give you a 1% more to trigger it, and each clover gives 0.5%~~
changed to be 25% chance to trigger luck at 100luck+50luck on rod
sidenote: putting clovers on actual accessories like bobbers and so on does nothing (so only upgrade the ROD), cant disable the ability to do so either yet as there is no way to "blacklist" items from tempering at the moment
fisheing

### DIFF
--- a/code/_core/obj/item/fishing/rod.dm
+++ b/code/_core/obj/item/fishing/rod.dm
@@ -189,9 +189,6 @@
 
 	INTERACT_CHECK
 
-	if(compact)
-		caller.to_chat(span("notice","Extend your fishing rod first!"))
-		return TRUE
 	if(object.plane >= PLANE_HUD)
 		return ..()
 
@@ -245,6 +242,9 @@
 			return TRUE
 
 	if(istype(object,/turf/simulated/hazard/))
+		if(compact)
+			caller.to_chat(span("notice","Extend your fishing rod first!"))
+			return TRUE
 		var/turf/simulated/hazard/H = object
 		if(!H.fishing_rewards)
 			return TRUE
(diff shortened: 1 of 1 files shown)


code/_core/obj/item/fishing/rod.dm
```diff
@@ -44,6 +44,13 @@
 	LOADATOM("line")
 	LOADATOM("bait")
 	
+/obj/item/fishing/rod/update_value()
```
BurgerLUA, line 47: This should be ``get_value()`` instead. Check out ``/obj/item/weapon/ranged/energy/get_value()`` on how that should work.

code/_core/obj/item/fishing/rod.dm
```diff
@@ -180,16 +189,19 @@
...
-	if(compact)
-		caller.to_chat(span("notice","Extend your fishing rod first!"))
-		return TRUE
 	if(object.plane >= PLANE_HUD)
 		return ..()
 
 	if(fishing_turf)
 		if(snagged_fish) //Caught something!
+			var/mob/living/C = caller // here comes the luck calc
+			var/lucktotal = clamp((C.get_attribute_level(ATTRIBUTE_LUCK) + (src.luck/2)- 75),1,75)	//base player+rod is 0 unless boosted, item luck affects it by half
```
BurgerLUA, line 198: This function already exists in its own form; ``luck()``. check out ``/proc/luck()`` on how this works.

BurgerLUA (changes requested):

code/_core/obj/item/fishing/rod.dm
AdrienTheMan: should be fine now possibly, take a look at the changes whenever

BurgerLUA (approved):

PR #182, opened 2021-03-13
Adds Pathfinding

BurgerLUA:
# What this PR does

Adds Burgerstar pathfinding.

Helps Peanut find things to kill as well as unstuck mobs who are stuck on a path.


BurgerLUA (commented):
i made a pr where more than half the code isn't actually implemented thanks to the scp being disabled

PR #189, opened 2021-03-15
Ghosts/observers can now "orbit" (follow) mobs

jjpark-kb:
# What this PR does
title

for my laziness.
when I am a ghost, sometimes I just want to watch and not have to move to follow them.

### DIFF
--- a/code/_core/client/verbs/view.dm
+++ b/code/_core/client/verbs/view.dm
@@ -79,6 +79,93 @@
 	if(is_living(mob))
 		log_admin("[src.get_debug_name()] jumped to [C.mob.get_debug_name()]'s location.")
 
+/client/verb/orbit_player()
+	set name = "Orbit Player"
+	set category = "View"
+
+	//sortTim(all_mobs_with_clients,/proc/cmp_path_asc)
+
+	var/mob/choice = input("Who would you like to orbit?","Orbit Mob") as null|mob in all_mobs_with_clients
+	if(!choice || choice == mob)
+		to_chat(span("warning","Invalid mob."))
+		return FALSE
+
+	var/turf/T = get_turf(choice)
+	if(!T)
+		to_chat(span("warning","Invalid turf."))
+		return FALSE
+
+	mob.force_move(T)
+	choice.observers += mob
+	mob.observed = choice
+
+	to_chat(span("notice","You orbit [choice]."))
+	if(is_living(mob))
+		log_admin("[src.get_debug_name()] chose to orbit [choice.get_debug_name()].")
+
+/client/verb/orbit_mob()
+	set name = "Orbit Mob"
+	set category = "View"
+
+	//sortTim(all_mobs,/proc/cmp_path_asc)
+
+	var/mob/choice = input("Who would you like to orbit?","Orbit Mob") as null|mob in all_mobs
+	if(!choice || choice == mob)
+		to_chat(span("warning","Invalid mob."))
+		return FALSE
+
+	var/turf/T = get_turf(choice)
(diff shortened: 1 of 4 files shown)


code/_core/mob/movement.dm
```diff
@@ -74,6 +74,13 @@
 		if(loc != old_loc)
 			post_move(old_loc)
 
+	if(length(observers))
```
BurgerLUA, line 77: You don't need to check the length of observers for a for loop as this is already done in the backend for the for loop itself.

code/_core/mob/movement.dm
```diff
@@ -74,6 +74,13 @@
...
+		for(var/mob/chosenObserver in observers)
```
BurgerLUA, line 78: Should do 
```
for(var/k in observers
 var/mob/chosenObserver = k
```
As it is cheaper.

code/_core/mob/movement.dm
```diff
@@ -56,7 +56,7 @@
 		. *= 2 - (health.health_current/health.health_max)
 
 
-/mob/Move(NewLoc,Dir=0,step_x=0,step_y=0)
+/mob/Move(NewLoc,Dir=0,step_x=0,step_y=0,var/control_move = TRUE)
```
BurgerLUA, line 59: I don't like adding control_move to this as there are other solutions. Maybe go into the ``chosenObserver`` loop and add 
```
if(chosenObserver.move_dir)
 observed.observers -= src
 observed = null
 continue
```

code/_core/mob/_mob.dm
```diff
@@ -93,6 +93,9 @@
 
 	var/last_z = 0
 
+	var/list/observers = list() //A list of observers/ghosts
```
BurgerLUA, line 96: I don't see this variable being .Cut() in Destroy() anywhere, so you should add that.

code/_core/mob/_mob.dm
```diff
@@ -93,6 +93,9 @@
...
+	var/mob/observed //who is being observed
```
BurgerLUA, line 97: Same here, this needs to be set to null in Destroy() for garbage deletion.

code/_core/client/verbs/view.dm
```diff
@@ -79,6 +79,93 @@
...
+	if(!T)
+		src.to_chat(span("warning","Invalid turf."))
+		return FALSE
+
+	mob.force_move(T)
+	C.mob.observers += mob
+	mob.observed = C.mob
+
+	to_chat(span("notice","You jumped to [C]'s location."))
+	if(is_living(mob))
```
BurgerLUA, line 166: I think it would be better, for now, to make it so that only observers can actually orbit players. There will be issues if living were able to orbit players.

BurgerLUA (changes requested):
good pr, just needs some changes

code/_core/mob/movement.dm
```diff
@@ -74,11 +74,11 @@
 		if(loc != old_loc)
 			post_move(old_loc)
 
-	if(length(observers))
-		for(var/mob/chosenObserver in observers)
-			chosenObserver.glide_size = glide_size
-			chosenObserver.Move(NewLoc, Dir, step_x, step_y, control_move = FALSE)
-	if(observed && control_move)
+	for(var/k in observers)
```
BurgerLUA, line 77: Actually I'm an idiot and should've suggested all of this should go in ``mob/post_move(var/atom/old_loc)`` and other things but I can do that myself.

BurgerLUA (approved):

PR #194, opened 2021-03-18
the blood totem

jjpark-kb:
# What this PR does
adds the blood totem, both regen and degen

more variety for totems, additional healing


code/_core/obj/structure/totem.dm
```diff
@@ -323,3 +323,47 @@
...
+			continue
+		if(L.loyalty_tag != affecting_faction)
+			continue
+		if(L.immortal)
+			continue
+		if(!istype(L.health))
+			continue
+		if(L.blood_volume + (5*leveled_effect) >= L.blood_volume_max)
+			continue
+		L.blood_volume += (5*leveled_effect)
```
BurgerLUA, line 346: I think you should instead do a check if the blood volume is greater or equal to the blood volume max, and then make it so that the amount of blood added is ```min(blood_volume_max - blood_volume_current,leveled_effect*5)```

code/_core/obj/structure/totem.dm
```diff
@@ -323,3 +323,47 @@
...
+			continue
+		if(L.loyalty_tag == affecting_faction)
+			continue
+		if(L.immortal)
+			continue
+		if(!istype(L.health))
+			continue
+		if(L.blood_volume - (5*leveled_effect) <= 0)
+			continue
+		L.blood_volume -= (5*leveled_effect)
```
BurgerLUA, line 368: Same here.

BurgerLUA (changes requested):

BurgerLUA (approved):

BurgerLUA:
Will be merged tomorrow.

BurgerLUA (approved):

PR #199, opened 2021-03-23
make station-side afk farming less rampant

jjpark-kb:
# What this PR does
default mobs, if they are station-side, they do not reward exp.

people are abusing stamina damage on mobs to reach really fast farming with no risk.
Afk farming (on the station) should be the dummies, which are slow, not some goblin.

### DIFF
--- a/code/_core/mob/living/experience.dm
+++ b/code/_core/mob/living/experience.dm
@@ -67,6 +67,9 @@
 /mob/living/get_xp_multiplier()
 	if(master)
 		return 0
+	var/turf/srcTurf = get_turf(src)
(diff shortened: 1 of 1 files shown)


ondrej008:
This would also remove station-side summon XP farming in the arena, but this is a good chance as I am currently leveling myself from 5 melee to 100 melee, I started 7 minutes ago and as of posting this comment I have gained 50 levels.

BurgerLUA:
I think a better solution would be that stamina damage does not count towards XP.

jjpark-kb:
since burger is addressing this issue, closing.

PR #206, opened 2021-03-28
Adds the 12.7x90mm browning deployable

Cyaneer:
# What this PR does

Adds more dakka.

NT Ops shop will have the deployment item.
After placing, insert the magazine, alt-click to take the handles, wield them, shoot in a 90 degree angle, hold the LZ. Can be picked up by dragging to the player.

Currently only for NT, might add other types later. Still needs balancing, probably.

More dakka more good. Allows for easy point defense.


code/_core/obj/item/deployable/_deployable.dm
```diff
@@ -8,8 +8,11 @@
 	item_count_max = 10
 	item_count_max_icon = 3
 
-/obj/item/deployable/proc/get_deploy_time(var/mob/caller)
-	return SECONDS_TO_DECISECONDS(3)
+/obj/item/deployable/proc/get_deploy_time(var/mob/caller, var/obj/item/deployable/D)
+	if(istype(D, /obj/item/deployable/mountable))
+		return SECONDS_TO_DECISECONDS(10)
+	else
+		return SECONDS_TO_DECISECONDS(3)
```
BurgerLUA, line 15: This shouldn't be setup like this. You should make a new proc like

```
/obj/item/deployable/mountable/get_deploy_time(var/mob/caller, var/obj/item/deployable/D)
 return SECONDS_TO_DECISECONDS(10)
```

BurgerLUA (changes requested):

code/_core/obj/structure/interactive/mountable/browning.dm
```diff
@@ -0,0 +1,297 @@
...
+	direction = src.dir
+
+	mag_check(src)
+
+//pack 'em up, from turret
+/obj/structure/interactive/mountable/browning/proc/can_pack_up(var/mob/caller)
+
+	INTERACT_CHECK_NO_DELAY(src)
+
+	if(get_dist(caller,src) > 1)
```
BurgerLUA, line 137: Distance is already checked in INTERACT_CHECK_NO_DELAY so you don't need this

code/_core/obj/structure/interactive/mountable/browning.dm
```diff
@@ -0,0 +1,297 @@
...
+
+	var/obj/item/deployable/mountable/browning/S = new(get_turf(src))
+	INITIALIZE(S)
+	FINALIZE(S)
+	if(stored_magazine)
+		S.stored_magazine = stored_magazine
+		stored_magazine.force_move(S)
+		stored_magazine = null
+
+	qdel(src)
```
BurgerLUA, line 159: For organization sake, qdel(src) should be called at the end of this proc.

code/_core/obj/structure/interactive/mountable/browning.dm
```diff
@@ -0,0 +1,297 @@
...
+
+			return TRUE
+
+	return ..()
+
+//taken from defib paddles
+/obj/item/browning_handle/click_on_object(var/mob/caller,var/atom/object,location,control,params)
+
+	if(is_inventory(object))
+		return FALSE
```
BurgerLUA, line 260: This should be ``return ..()``

BurgerLUA (changes requested):

code/_core/obj/structure/interactive/mountable/browning.dm
```diff
@@ -0,0 +1,297 @@
...
+/obj/structure/interactive/mountable/browning/Finalize()
+
+	stored_weapon = new stored_weapon(src)
+	INITIALIZE(stored_weapon)
+	GENERATE(stored_weapon)
+	FINALIZE(stored_weapon)
+
+	stored_weapon.linked_gun = src
+
+	stored_handle = new /obj/item/browning_handle(src)
```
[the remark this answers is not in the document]
Cyaneer: CREATE()'s defined as this:
#define CREATE(I,desired_loc) var/datum/D = new I(desired_loc);INITIALIZE(D);GENERATE(D);FINALIZE(D)
Are you sure it can make /obj/ects? I'm getting tons of undefined vars.

code/_core/obj/item/deployable/_deployable.dm
Cyaneer: That gives a duplicate definition, and would a need a separate on_click deploy for mountables or an if() in the deployable on_click instead. Do I separate mountable code from _deployable then?

BurgerLUA:
Is this good to merge?

Cyaneer:
Not yet, I'll see if I can make it work with the vardels.

Cyaneer:
Changes are in, gun finally works, nothing seems to be breaking.

Cyaneer:
PR should be finished for now.

BurgerLUA (approved):

PR #223, opened 2021-04-27
generator station power

jjpark-kb:
# What this PR does
adds a generator(engine) to the station.
if the station does not have power, doors won't open (without crowbars). Vendors won't work, and the shuttle launcher won't work stationside (still will work planetside).
if the engine goes all the way down, lights will go off until the engine is fully repaired

more stationside jobs (that are critical)


code/_core/obj/structure/interactive/door/airlock/_airlock.dm
```diff
@@ -317,6 +320,18 @@ obj/structure/interactive/door/airlock/close(var/mob/caller,var/lock = FALSE,var
...
+	INTERACT_CHECK
+	INTERACT_CHECK_OBJECT
+	INTERACT_DELAY(1)
+
+	if(is_item(object))
+		var/obj/item/objI = object
+		if(objI.flags_tool & FLAG_TOOL_CROWBAR && door_state == DOOR_STATE_CLOSED)
+			open(caller, FALSE, TRUE)
+
+	return TRUE
```
BurgerLUA, line 334: Does this work properly with empty hand clicking? Currently you can open airlocks by clicking on them but I'm wondering if this code overwrites that behavior.

code/_core/obj/structure/interactive/fusion_engine.dm
```diff
@@ -0,0 +1,85 @@
...
+/obj/structure/interactive/fusion_generator/proc/tool_repair(mob/caller)
+	caller.to_chat(span("notice", "You finish repairing the generator."))
+	cooldown_time = world.time + SECONDS_TO_DECISECONDS(add_cooldown)
+	stage++
+	icon_state = "gen_small_[stage]"
+	update_icon()
+
+/obj/structure/interactive/fusion_generator/get_examine_list(mob/examiner)
+	. = ..()
+	. += div("notice","Current Stage: [stage]")
```
BurgerLUA, line 76: This should probably be removed. Seems like a debug message.

code/_core/obj/structure/interactive/fusion_engine.dm
```diff
@@ -0,0 +1,85 @@
...
+	cooldown_time = world.time + SECONDS_TO_DECISECONDS(add_cooldown)
+	if(prob(50)) //50 percent chance to not degrade!
+		return TRUE
+	stage--
+	icon_state = "gen_small_[stage]"
+	update_icon()
+	if(stage <= 0) //if stage is 0
+		SSgenerator_power.station_power = FALSE
+		SSgenerator_power.turn_off_lights()
+	return TRUE
```
BurgerLUA, line 42: I think you should incorporate ``return FALSE`` here as this will go into the negatives in terms of stages. Also it's better for performance if it doesn't think.

But I think it would be better if you made a new proc with this loop called /power_think and move the thinking to the new subsystem you made, and make the new subsystem think every 5 minutes or however often you want the system to reset.

code/_core/obj/structure/interactive/light_switch.dm
```diff
@@ -50,6 +50,9 @@
 	return ..()
 
 /obj/structure/interactive/light_switch/proc/toggle(var/mob/caller)
+	if(!SSgenerator_power.station_power)
+		play_sound('sound/machines/click.ogg',get_turf(src),range_max=VIEW_RANGE*0.5)
+		return TRUE
```
BurgerLUA, line 55: I don't know how hard this will be, which is why I'll probably do it if you're not able to, but I think the lightswitch should switch its on-off state but the lights itself don't actually update.

code/_core/obj/structure/interactive/lighting/fixture/_fixture.dm
```diff
@@ -37,10 +37,14 @@
 	else
 		create_destruction(T,list(/obj/item/material/sheet = 1),/material/steel)
 		. = ..()
+		if(src in SSgenerator_power.stationside_lights)
+			SSgenerator_power.stationside_lights -= src
```
BurgerLUA, line 41: You don't need to check if src is in the stationside lights. It's safe, faster, and better to just remove it even though it might not exist in the first place (unless you're doing special behavior after it).

code/_core/world/subsystems/generator_power.dm
```diff
@@ -0,0 +1,23 @@
+SUBSYSTEM_DEF(generator_power)
+    name = "Generator Power Subsystem"
+    desc = "The system for turning off/on lights, and deciding if there is station power."
+    priority = SS_ORDER_NORMAL
+
+    var/list/stationside_lights = list()
+    var/station_power = TRUE
+
+/subsystem/generator_power/proc/turn_off_lights()
+    for(var/obj/structure/interactive/lighting/fixture/L in stationside_lights)
```
BurgerLUA, line 10: You should turn this into a for(var/k) loop, and then make it so that L = k. For checks like this intrinsically run istype() while looping through a list, which is redundant if your list contains only one type of object, which is a light fixture.

BurgerLUA (changes requested):
Amazing PR. Just need to change a few things and it should be good to merge.

code/_core/obj/structure/interactive/fusion_engine.dm
jjpark-kb: I could remove it, but its really for flavor and can help people know how much longer they have until power cuts.

code/_core/obj/structure/interactive/light_switch.dm
jjpark-kb: I'm not sure I quite understand.

code/_core/obj/structure/interactive/door/airlock/_airlock.dm
jjpark-kb: fixed. I can open and close airlocks. I can use a crowbar to open the airlocks when unpowered and close them with my hand.

jjpark-kb:
closing this for now. Will eventually PR it back up once everything is sorted.

PR #227, opened 2021-05-31
prescription for mentally deranged players (bags of hoarding)

AdrienTheMan:
# What this PR does
-adds a special "bag" which can store unlimited amounts of a single special currency (TC, magic tokens, antag tokens, coins, prize tickets)
once a currency has been put in, it will be locked to it until emptied, you can also choose how much you can take out anytime, sold at the costume vendor next to bar as a placeholder location _but it kinda works_
-makes some players stop having bags full of magic token burgers making the saves a little lighter (: maybe
-also gives some spaghetti for some hungry hungry coders even if it functions properly (this was a mess to make)
sex


ChewedPencil:
pog

ilikesaying:
nanospores

code/_core/obj/item/storage/bags/_bags.dm
```diff
@@ -327,4 +327,124 @@
...
+
+/obj/item/storage/bagofhoarding
+	name = "bag of hoarding"
+	desc = "we just warframe now"
+	desc_extended = "A peculiar bag that can hold unlimited amounts of specific rare currencies. Can be dyed."
+	icon = 'icons/obj/item/storage/boxes.dmi'
+	icon_state = "moneybag"
+	value = 10000
+	var/hoard = 0				// how many inside
+	var/fuck					// this one is used for the base currency roots, look further down for comments
```
BurgerLUA, line 340: Just a minor nitpick, the variable should be somewhat straightforward without comments. It just makes it easier for me and future contributors to understand.

BurgerLUA (changes requested):

code/_core/obj/item/storage/bags/_bags.dm
AdrienTheMan: FAIR ENOUGH i set it that due to the mental damage makin this gave me, should be fixed now plus i got to squash a save-bug at the same time

BurgerLUA (approved):

PR #276, opened 2021-07-11
QOL shit 1 [Ready!]

Moltovz:
This does a lot...
# What this PR does
- Adds chests to almost all ruins
- Makes the mining drop pod good again
- removes errors from the rev compound
- Makes *antag tokens* a contract reward (bosses only) ((The description doesn't properly say rewards.... sad!))
- Adds more enemies to Deep jungle
- Adds IRF scanners to the lower side of the station
# Shit looks good on a private server. Test the hell out of it, you have time since the server is down.
More importantly:
- Drip now drops from the gobbo king very rarely.


code/_core/turf/unsimulated/generation/jungle_deep.dm
```diff
@@ -52,16 +52,38 @@
...
+				new /obj/marker/generation/jungle_tree_small(src)
+			else if(prob(1))
+				new /obj/marker/generation/jungle_high_grass/small(src)
+			if(prob(1))
 				new /obj/marker/generation/jungle_light_stick(src)
 			else if(prob(1))
+				new /obj/marker/generation/mob/arachnid(src)
+			else if(prob(1))
+				new /obj/marker/generation/jungle_high_grass/small(src)
+			if(prob(1))
```
BurgerLUA, line 64: In some instances you're breaking the else if chain. You should be careful with this and make sure you mean else if and not just if as it may cause objects to spawn inside eachother.

BurgerLUA (changes requested):

code/_core/turf/unsimulated/generation/jungle_deep.dm
Moltovz: For whatever reason when they *were* else if the mobs wouldnt spawn at all, OR trees/plants in the area wouldn't spawn.

Moltovz: image
Im not saying items *dont* spawn in eachother, but you seriously need to go far out of your way to find them

BurgerLUA (approved):

PR #279, opened 2021-07-11
adding Botany enhancements.

drakewill-CRL:
# What this PR does
* Adds a Botany skill that increases the potency and yield of harvested plants. Botany skill is increased by harvesting plants.
* Adds a consumable Fertilizer item to the botany vendor to increase potency of plants for money. Plant-2-Grow adds 1 potency to 1 plant for 50 credits. 

A: Botany becomes more fun.
B: Botany becomes useful as a repeated action, so people that spend a lot of time growing plants can get the amount of resources they want faster. 
C: Botany primarily interacts with cooking and chemistry, so this provides some incentive to to mess around with those systems as well.
D: Fertilizer can act as a credit sink for those uninterested in leveling up Botany. 50 credits for Plant-2-Grow replaces 5 points in the Botany skill for a single plant, so you can spend 1,000 credits to get half the benefit of maxing out Botany for one plant (you won't get the yield increase). This is viable if you want a small number of plants for a specific purpose. 
E: All those random plants in the game world become worth walking over to and harvesting, since now they're a small XP boost to a secondary skill.
F: It's kinda fun making gigantic plants. Harvesting tomatoes bigger than you is cool to see. Gigantic plants can also make gigantic burgers, which 100% belong on burgerstation.


code/_core/datum/loot/scroll.dm
```diff
@@ -14,7 +14,8 @@
 		/obj/item/experience_scroll/skill/ranged,
 		/obj/item/experience_scroll/skill/sorcery,
 		/obj/item/experience_scroll/skill/survival,
-		/obj/item/experience_scroll/skill/unarmed
+		/obj/item/experience_scroll/skill/unarmed,
+		/obj/item/experience_scroll/skill/botany,
```
AdrienTheMan, line 18: one comma too much after botany

AdrienTheMan (changes requested):
syntax is pain, evil little commas

Stalkeros2:
Based PR.

code/__defines/experience.dm
```diff
@@ -48,4 +48,6 @@
 #define SKILL_MAGIC_OFFENSIVE "offensive" //Completed
 #define SKILL_MAGIC_SUMMONING "summoning" //Completed
 #define SKILL_MAGIC_ENCHANTING "enchanting" //Completed.
-#define SKILL_MEDICINE "medicine" //Completed
\ No newline at end of file
+#define SKILL_MEDICINE "medicine" //Completed
+
```
Stalkeros2, line 52: Rather certain that this line skip is unnecesary.

code/_core/datum/experience/skills.dm
```diff
@@ -140,4 +140,13 @@
 	desc = "This doesn't work, yet!"
 	desc_extended = "Your skill in enchanting objects. Affects the speed, sucess rate, and quality of enchanted items."
 	experience_power = 1.8
-	experience_multiplier = 4.1
\ No newline at end of file
+	experience_multiplier = 4.1
+
+/experience/skill/botany/ //Int
```
Stalkeros2, line 145: I know that this is gonna sound more of a "me" thing, but I really think that botany should use your strength or endurance-like stat, considering how much work do all of the farmers have to put into crops.

drakewill-CRL: Upon digging, this was an old comment from the previous iteration. As it exists in the current code, Botany skill doesn't actually rely on any specific stat. Updating comments to reflect this.

code/__defines/experience.dm
drakewill-CRL: It is helpful, since Botany isn't attached to a stat the way the existing 15 skills are organized. I would keep Botany in its own no skill attached category until there's 2 more skills, so the STR/DEX/INT skill symmetry can be retained. Comments will be added to clarify this.

code/_core/datum/loot/scroll.dm
drakewill-CRL: MORE COMMAS MORE GOOD
jk, removed.

code/_core/obj/item/consumable/fertilizer.dm
```diff
@@ -0,0 +1,15 @@
+/obj/item/consumable/fertilizer
+	name = "Plant-2-Grow"
+	desc = "Weird space fertilizer"
+	desc_extended = "Apply to a plant to increase its potency."
+	icon = 'icons/obj/item/container/spray_bottle.dmi'
+	icon_state = "inventory"
+	value = 50
+
+/obj/item/consumable/fertilizer/click_on_object(var/mob/caller as mob,var/obj/structure/interactive/plant/O,location,control,params)
```
BurgerLUA, line 9: click_on_object should have its aug types the same as the others, and the new variable type O should be created after the typecheck. This is to prevent any errors in the future and make all the other procs up to standard.

code/_core/obj/structure/interactive/botany/_plant.dm
```diff
@@ -96,6 +97,12 @@
 		if(move_direction & WEST)
 			animation_offset_x += 32
 
+		//Harvester's botany skill increases the harvested results based on their skill power.
+		//Each point in skill power add .2 potency and .1 yield.
+		var/skillPower = caller.get_skill_power(SKILL_BOTANY)
+		potency = potency + (potency * skillPower * 2)
+		yield += yield * skillPower
```
BurgerLUA, line 104: Not sure if this is intended, but with each harvest, the yield and potency of the plant would increase. This wouldn't just affect the current harvest, but future harvests as well.

BurgerLUA (changes requested):

code/_core/obj/structure/interactive/botany/_plant.dm
drakewill-CRL: I'll update the after-harvest code to reset these back to the initial values, instead of (initial + current)/2

BurgerLUA (approved):

PR #280, opened 2021-07-11
Enable magazine enhancements

drakewill-CRL:
# What this PR does
This code addition allows magazines for weapons to be enhanced in 2 separate ways, and creates items to apply those enhancements.
* A Bluespacer enhancement allows a magazine to hold 10x its normal capacity.
* A Refiller enchantment lets a magazine regenerate 1 bullet every 3 seconds on it's own.
Both items are set to a cost of 50,000 credits. These would make a great credit sink, since combined a player would be dropping 100k credits on a single magazine (and therefore a single gun) as a personal goal or a way to escape managing a dozen reloads per trip out of the station. 

# What this PR DOESN'T do
* Make these enhancement items available. 
* Handle the weapons that don't use a magazine (lasers, revolvers, most shotguns)
* Add icons for these items. Currently reusing the whetstone icon until these get approved, because sprites are hard.
* Understand any balance concerns that these would break. I would love to discuss these somewhere
** I don't see any public list of rules/math that explain out 'balance' in Burgerstation, or any 'goals' for players beyond the core gameplay loop of killing bosses and restocking at base.
** I had originally assumed that these would be purchased from cargo or a vendor in the rich-player vending machines, hence the cost. If these were instead made as a boss drop, they won't dramatically inflate the economy since the conversion to gold means a player will only get 2400 credits from selling one. This should certainly encourage using or trading them over selling them to the bank NPC.
** I made these changes shortly before the first iteration of BurgerStation shut down, when the general assumption was that magic was overpowered, guns were meh, and melee was bad. On the short time the new iteration has been up, the new assumption seems to be that guns are good, melee is good, and magic is bad. An understanding of the distinction between the 3 categories and if there's any specific/unique properties that should be done to each/any of them would help work this into the game. Are they only separated by damage types available? Costs for items? Should there be something like these items for the other categories of attacks?

A: The core idea for this change is good for players and the station economy.
B: Players have a way to get out of one aspect of inventory management at a high (but viable) cost.
C: Players can have a goal to make their gun the best it can be, instead of just buying one from a vendor and running with it.
D: The goal is repeatable as players experiment and take a liking to different weapons, or occasionally die and lose their gun with the enhanced magazine.
E: I'm entirely open to discussions/adjustments of the idea and the numbers involved. Or making multiple steps of each if desired.

### DIFF
--- a/code/_core/obj/item/magazine/_magazine.dm
+++ b/code/_core/obj/item/magazine/_magazine.dm
@@ -21,8 +21,30 @@
 	var/bullet_diameter_max = -1
 
 	var/icon_states = 1
+	var/bluespaced = FALSE
+	var/regenerate = FALSE
+	var/regen_counter = 0
 
 	weight = 0.25
+
+
+/obj/item/magazine/PostInitialize()
+	start_thinking(src)
+	return ..()
+
+/obj/item/magazine/think()
+	if (regenerate)
+		regen_counter += 1
+		if (regen_counter > 30)
+			regen_counter = 0
+			if (length(stored_bullets) < bullet_count_max)
+				var/obj/item/bullet_cartridge/B = new ammo(src)
+				INITIALIZE(B)
+				FINALIZE(B)
+				stored_bullets += B
+				update_sprite()
+
+	return ..()
 
 
 /obj/item/magazine/update_icon()
@@ -50,6 +72,8 @@
 		for(var/i=1,i<=length(stored_bullets),i++)
 			var/obj/item/bullet_cartridge/B = stored_bullets[i]
 			if(B) .["stored_bullets"][B.type] += 1
+	.["bluespaced"] = bluespaced
+	.["regenerate"] = regenerate
 
 
 /obj/item/magazine/load_item_data_post(var/mob/living/advanced/player/P,var/list/object_data)
@@ -64,6 +88,13 @@
 				INITIALIZE(B)
 				FINALIZE(B)
 				stored_bullets += B
+
+	if (object_data["bluespaced"])
+		bullet_count_max *= 10
+		bluespaced = TRUE
+
+	if (object_data["regenerate"])
+		regenerate = TRUE
 
 
 /obj/item/magazine/Generate()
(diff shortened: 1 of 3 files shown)


Moltovz:
10X is a lot of ammo in one mag.

ilikesaying:
2000 shots DMCA

code/_core/obj/item/tempering/tempering_magazine.dm
```diff
@@ -0,0 +1,53 @@
...
+	desc = "Have you tried just putting more bullets in it?"
+	desc_extended = "A tightly controlled bluespace pocket condenser. Fits to expand inside a magazine, and increases capacity 10x for that magazine"
+	icon_state = "quality_melee"
+
+	temper_whitelist = /obj/item/magazine
+
+	value = 50000
+
+/obj/item/tempering/magazine/bluespace/can_temper(var/mob/caller,var/obj/item/magazine/I)
+	if (is_magazine(I) == FALSE)
```
BurgerLUA, line 17: ``x == FALSE`` should be converted into ``!x`` just in case something goes wrong. In byond, ! means not.

code/_core/obj/item/tempering/tempering_magazine.dm
```diff
@@ -0,0 +1,53 @@
...
+		return FALSE
+
+	if(I.bluespaced == TRUE)
```
BurgerLUA, line 20: ``x == TRUE`` should be converted into just ``x`` as it's the same thing, but neater.

code/_core/obj/item/magazine/_magazine.dm
```diff
@@ -21,10 +21,32 @@
...
+
+/obj/item/magazine/think()
+	if (regenerate)
+		regen_counter += 1
+		if (regen_counter > 30)
+			regen_counter = 0
+			if (length(stored_bullets) < bullet_count_max)
+				var/obj/item/bullet_cartridge/B = new ammo(src)
+				INITIALIZE(B)
+				FINALIZE(B)
```
BurgerLUA, line 43: GENERATE(B) should be placed between INITIALIZE(B) and FINALIZE(B) as an object is being generated for the first time.

code/_core/obj/item/magazine/_magazine.dm
```diff
@@ -21,10 +21,32 @@
...
+	var/regen_counter = 0
 
 	weight = 0.25
 
 
+/obj/item/magazine/PostInitialize()
+	start_thinking(src)
+	return ..()
+
+/obj/item/magazine/think()
```
BurgerLUA, line 35: I'm not sure I like how this is all setup because start_thinking() would be called regardless if it actually needs to think or not, and the start_thinking() should really be used for things that should be thinking really fast. What you should do instead is use the CALLBACK() define instead in a loop. ``/obj/structure/interactive/blob/proc/check_mobs()`` in _blob.dm is a good example on how to do this.


I'm aware that some other objects coded by me are coded like this, but that's bad code and I wish to change it in the future.

BurgerLUA (changes requested):

BurgerLUA (approved):

PR #289, opened 2021-07-15
Map tweaks

Moltovz:
# What this PR does
- Adds departmental pets
- Fixes up Mr. Scratch's office
- Three basic Mystery crates are available to purchase from the goblin merchant via tables in his room. (3,000 credits base each)
- A tiny arcade was added along with a prize vendor and random vendor.
- Empty spots in maint now have loot where there wasn't any before.
- Puts an enchanting vendor with the magic vendor as a placeholder (no chalk or tomes)


BurgerLUA (approved):
The animals should be their own subtype and this PR should've been more atomized, but to prevent development hell I'll just merge it and change everything I don't like later.

PR #291, opened 2021-07-16
Adds aluminium to smelter (I FORGOT PLS SPEEDMERGE)

Stalkeros2:
# What this PR does
Adds aluminium smelting

Makes aluminium smeltable and that's good because it's used in, like, half of the available recipes.


Stalkeros2:
Obsolete due to #293

PR #294, opened 2021-07-16
food bags

Stalkeros2:
# What this PR does
Adds food bags to the kitchen vendor.

Food tasty and good and you can also carry it around in bulk now.


BurgerLUA:
Not sure I actually like this. It seems really powerful and just makes our already existing paper bags kinda useless.

Stalkeros2:
Oh shit, I forgot oaper bags exist.

FancyCyclops:
paper bags are real, and can hurt you

so use them to acquire your meats instead

PR #297, opened 2021-07-16
let banks hold empty size 3+ containers

### LINKED ISSUE #296: containers cant stop being containers
if you have a container, say a bag or paper bag or whatever, and you change its dynamic_inventory_count to 0 or is_container to false, you can still open them no matter what (which did not function this way before)

AdrienTheMan:
# What this PR does
-allows the bank to hold any container that is 3+ in size if they are empty, but also makes them unable to contain items until taken out of the bank
-yea
-idk thats it
(:

### **FIX #296 BEFORE MERGE** 
or else people will still be able to use them when banked

sidenote: cant find a way to remove specific "close inventory" buttons, its all of them or nothing, could there be something done for that?

the


BurgerLUA:
Closing this as I will do my own implementation of this, or at least try to.

PR #307, opened 2021-07-17
adds proof of concept cosmetic headset

FancyCyclops:
Hacky code, mistake made.

May god forgive my transgression, for I know many on this earth won't.

# What this PR does

adds cat ear headset that goes over everything as a proof of concept for unique looking headsets

hopefully does not make them obtainable

It shouldn't. But I'm going to make this request anyway and see what happens.
because it'll be funny.


Stalkeros2 (approved):
Not based but also based.

PR #310, opened 2021-07-18
Makes magician's gems work, for real.

FancyCyclops:
# What this PR does
Acts as a bandaid to the problem and makes magician's gems work on spellgems.

until magic is finalized, at least let us have this...

image


SarmentiCampbell:
what about the book that do aoe fire damages

Stalkeros2 (approved):
once the suppression field is down, i'm going to mate

PR #312, opened 2021-07-19
Quality of Life Changes

drakewill-CRL:
# What this PR does
* Adds more items to the belt quickslots
** Mags, batteries, cleaning kit, and medical items will go to an empty hand when used from a quickslot
** Health/value/armor/genetic scanners will scan the clicked item when used from a quickslot.
** Pinpointers can be set from the belt directly
* Adjusts the pricing of odd-typed items to match reasonably good guns. They're less cash than a crate now.
* Ranged weapons with multiple firing modes now tell you how many, and how to change them
* Wishgranters now have a sound effect if you get loot from them. 

Better belts, -2 easy cashgrabs, better descriptions, "DAH DAAAAAAAAAAAAH" when you get loot!


Stalkeros2 (approved):
All those PRs are based.

PR #315, opened 2021-07-20
Syndicate Marskman rifle buff

NzzyFk:
# What this PR does
"Buffs" the .308 AR-M
Changes the RPM and accuracy of the AR-M. 
Currently, the .308 AR-M doesn't have automatic fire and is weaker than the SOL-15, leaving it feeling kinda shit.


Stalkeros2 (approved):
Looks more like a rebalance rather than a buff but let's see how this works.

PR #316, opened 2021-07-21
Honey Badger PDW

NzzyFk:
# What this PR does
Adds the .300 Mead Squirrel to the syndicate arsenal

Revs have the AS Val,so syndies should also get a stealth gun.


Stalkeros2 (approved):
Based

PR #321, opened 2021-07-21
Finally blundergauss

FancyCyclops:
# What this PR does
adds the Blundergauss, which is the Gauss Rifle, but break-action and TINY

you know what we need?
more russian content


Stalkeros2 (approved):
Based

PR #323, opened 2021-07-21
hopefully fixes slaughter_heavy

FancyCyclops:
so my fix broke something but at the same time fixed something

# What this PR does
slaughter/heavy is now slaughter_heavy

eating corpses is for non-harmful intents!!!!

...I just don't want people staying on unga-harm, okay?


Stalkeros2 (approved):
My tummy hurts.

PR #331, opened 2021-07-24
Adds the hypospray.

jc-denthead:
I got the sprites [from here,](https://github.com/yogstation13/Yogstation/blob/c2f0f7f380a680ba29455809ed942d9b443d17f0/icons/obj/syringe.dmi) but it comes from TG. **As far as I've checked, this works.**

# What this PR does
Adds a **60u** medipen that can switch between **10u** and **20u** injections.

This also adds the hypospray to syndicate scientists and the syndicate vendor for around 4TC.

Medipens deserve more love, so why not add an upgraded version for syndies? This isn't at all to sate my crippling addiction to medipens. I swear.


Stalkeros2 (approved):
I forgor :skull:

PR #342, opened 2021-07-31
Ballistic Shield Sprite Fix + Blundergauss Rebalance

NzzyFk:
# What this PR does
fixes ballisticshield  directional sprites
rebalances blundergauss inaccuracy from 3 to 2 so it can actually hit

Because I want my swat drip to work and also because blundergauss deserves love


Stalkeros2 (approved):
merging in unblocked

PR #345, opened 2021-08-01
hypos can now inject in 10u increments

jc-denthead:
# What this PR does
**Works as far as I tested it.**

This PR blatantly steals code from drinking glasses, so now you can inject in 10u increments.

More freedom for injecting with hyposprays is pretty good,


Stalkeros2 (approved):
ниг

PR #346, opened 2021-08-01
Increased maximum amount of coins in the stack

Stalkeros2:
# What this PR does
Increases the maximum amount of coins in the stack from 200 coins to 4k coins.

Better storaging good.


BurgerLUA:
I think the 200 coin limit is fine. It prevents gold hoarding and farming to a reasonable amount.

ilikesaying:
meanwhile every veteran player is walking around with the bag of hoarding

PR #347, opened 2021-08-01
Ore var adjusment

NzzyFk:
# What this PR does
more ore?? i think?? maybe??????

fun??


Stalkeros2 (approved):
Hopefully this does something.

PR #349, opened 2021-08-02
Adds Oxygen Adjustment to Slaughterclaws + changes bloodcrawl sound

FancyCyclops:
# What this PR does
Whenever the Slaughterclaws would change blood, they also adjust your oxyloss by the same amount.
also changes the jaunt effect to something more bloody, but it's a bit quieter. May undo change if controversial.

Less confusion, more feedback.


Stalkeros2 (approved):
eeee

PR #350, opened 2021-08-02
Bumps Flamer Damage Up

FancyCyclops:
# What this PR does
WAAAAAAAH I CAN'T KILL ANYTHING WITH THIS THING!

...seriously. I can't. And bumping the fire rate up might be a bad idea, so...

why even use this?

(TO DO: STORAGE CAPABILITIES WHEN?)


BurgerLUA:
I was told that this weapon is extremely powerful.

BurgerLUA:
Will be taking a hard look at the flamer and nerfing/buffing it myself.

PR #356, opened 2021-08-05
QoL stuff and Botany Re-Re-Fixed

drakewill-CRL:
# What this PR does
New QoL Patch:
* Portal scrolls are now usable from the belt pouch!
* GPS devices now save info between rounds! You'll have to set any existing ones you owned, since they didn't save data before.
* Bluespacers are consistent again.
* Filters now say in their description how to set their filter value!
* **BOTANY UN-FUCKED! WAIT HOLD ON LET ME EXPLAIN:**
** Plants grow to full again before dying of old age!
** You get more than 1 XP for havesting plants! 
** Your Botany skill once again impacts the yield and potency of plants!
** Taking good care of plants can undo damage they received!
** Plant scanner added so you can figure out what's going on with growing plants now!
** Corn yield halved. Just, the single most obvious fix to corn being good enough to build a country on.
** BUT
** Plants take damage if they're not in the right hydration/nutrition range.
** Damage affects the harvest, so injured plants will give fewer/smaller crops.
** Dead plants automatically disappear, so you might lose a harvest if you can't get to them in time.
** Crappy soil makes plants grow slower. Way slower. You want to buy that plant-2-grow to make growing plants work.
** Oh, if botany eats too much CPU time, EVERY PLANT GETS WIPED OFF THE SERVER.

NEW BOTANY CHEAT SHEET:
Plants need 1 water for every 2 seconds they're alive, and 1 nutrition for every 8.
Plants gain between 3-8 growth every second (varies by crop) and they need 100 to harvest.
Plants slowly soak in water/nutrients, so it takes a few seconds for the description to change after adding water/plant food.
Single-harvest crops can be planted without much worry, as long as you're there waiting for them.

Belts get better, GPS does the stuff I COULD HAVE SWORN I DID LAST WEEK I MUST HAVE DREAMED THAT CODE UP I GUESS, and botany is a thing people will do without crashing the server and soloing the skill 0-100 in 30 minutes.


code/_core/obj/item/tempering/tempering_magazine.dm
```diff
@@ -30,7 +30,7 @@
 
 /obj/item/tempering/magazine/bluespace/on_temper(var/mob/caller,var/obj/item/magazine/I)
 	I.bluespaced = TRUE
-	I.bullet_count_max *= 4
+	I.bullet_count_max *= 10 //Inconsistency leaves a sad taste in my mouth
```
BurgerLUA, line 33: Whats the reason for this change?

BurgerLUA (dismissed):
Very good PR

code/_core/obj/item/tempering/tempering_magazine.dm
drakewill-CRL: To make the bluespacers consistent. Description says 10x, on loading a saved item its 10x, but when you use one for the first time, its 4x because of this line. Saving/loading your character immediately after makes it go back to 10x. This is the smallest, quickest change to get people to stop asking/complaining that the 10x enhancement only adds 4x. 

Now that the items have been in use for a couple weekends, I think that 10x is a good number for getting people's attention and makes the 55k price worth it, but that it might be better balance-wise to have the bluespacer check for specific mags and adjust the amount the size is increased, and then save/load the actual capacity instead of (or in addition to) just the 'bluespaced' flag to 10x the capacity. Main example: the DMCA breaks from hitting -100% quality long before actually using all 10x of its bullets with this enhancement. Some guns get much more value out of the regen enhancement than the bluespacer like the tranq pistol (which is fine overall, though a good example of where no amount of multiplier makes this enhancement really worth the price). Some automatics are fine with the extra capacity being 10x (SMP seems OK with 10x without being super-jank) and some seem to have relied heavily on those couple seconds of reloading to stay roughly balanced (20g Bulldog can cut down forests about as fast as most bosses with this). 

Setting up a full list of which mags should get how much extra capacity and the code to handle that is a bigger change than I could get done before the next test. This fix is mostly to stop people from asking 'are bluespacers broken? i only got 4x on my 10x mag boost' while that gets figured out.  I'm also OK with another user taking the lead on figuring out which ones should get what multiplier for balance purposes, since I am quickly developing a reputation for favoring 'big numbers fun' over 'fair numbers good'

BurgerLUA (approved):

PR #373, opened 2021-08-08
Where I rip out flamethrowers from fed

FancyCyclops:
# What this PR does
makes flamethrowers infinite ammo, justifying its place as a back-slot item in a lazy way
oh yeah, the low charge cost SHOULD mean that it gains shots back reasonably fast

It could use some work.


Stalkeros2:
Should be a better alternative to a flamethrower that can be recharged with anything.

Stalkeros2 (approved):
amogus

PR #374, opened 2021-08-08
Holy M1911 Nerf

NzzyFk:
# What this PR does
makes the holy m1911 not ultra grind your levels

because grind


NzzyFk:
the gun is too blessed

FancyCyclops:
-Why not just set the gun's bonus to 0?

Issue #340, opened 2021-07-31
Rogue pixels BFR rifle in holster slot


Degirin2120:
The BFR Sniper Rifle, when holstered, has a single white pixel on the north/south facing side sprites.

PR #376, opened 2021-08-09
Fixes a rogue pixel on the Syndicate .50 rifle

SylvinSeik:
# What this PR does
Fixes the holstered rogue pixel as per PR title

Rogue pixels make me seethe.
Closes issue #340 
I hope that link works I always wanted to do that the fancy way.


Stalkeros2 (approved):
am
ogus

PR #382, opened 2021-08-09
Crafting ID

NzzyFk:
# What this PR does
sets crafting id to a bunch of shit

so i can make hacksaw recipes


BurgerLUA:
This is a little crazy and I unironically regret not stopping you sooner.

crafting_id is meant to assign an ID to a bunch of objects that would share the same ID. This is not a good use of crafting_id as just getting its type is better instead.

BurgerLUA:
closing this as this PR isn't needed, unfortunately.

PR #384, opened 2021-08-11
fixes fultons not disappearing

### LINKED ISSUE #370: Fultons don't properly leave
A bit of a uncommonish bug, but sometimes when fultoning a mob/object, the fultoned item won't disappear, taking up the tile.

ghost:
# What this PR does
Resolves #370. As far as I tested it, this works.
It's really annoying needing to click under an invisible object. If you fultoned something in just the right place, a chest or corpse could be made inaccessible.


BurgerLUA (approved):
can't believe I forgot the qdel

PR #388, opened 2021-08-14
Adds CURSED COIN to the CONTRACTOR, affects luck

Harrivchaw:
# What this PR does
see title
has a chance to increase/decrease your luck by 1 level increments
fleshes out NPC

### DIFF
--- a/code/_core/obj/item/coin/_coin.dm
+++ b/code/_core/obj/item/coin/_coin.dm
@@ -9,6 +9,39 @@
 
 	weight = 0.05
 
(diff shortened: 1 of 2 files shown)


BurgerLUA:
I'm a little conflicted about something that increases luck like this that can directly be purchasable.

Harrivchaw:
should I make this a boss drop? fishing drop? increase the price? make it so the odds be mostly neutral (i.e. 75% 0, 12.5% +1, 12.5% -1)?

PR #416, opened 2021-08-24
Domni Dextera!

ChewedPencil:
# What this PR does
adds a prayer unarmed weapon to pummel your foes in the name of God!

cause more unarmed types for different damage types


ChewedPencil:
Moving changes to Courier Coats cause i fucked up

PR #432, opened 2021-09-09
Adds a bunch of icons

aGhostCoder:
# What this PR does
Adds icons

I don't know.

# Notes
They aren't properly coded at all, also I haven't used github in awhile so I kind of forgot how to do this. apologies if it reopens the previous pull.


BurgerLUA (changes requested):
Not sure what the issue is but the two files above are conflicting.

BurgerLUA (approved):

PR #456, opened 2022-02-21
Deagle Resprite

NzzyFk:
# What this PR does
Changes the deagle in-hand and item sprite

So it feels cooler to use.


Stalkeros2 (approved):
I agree

PR #457, opened 2022-03-02
Syndibase Prefab

Moltovz:
# What this PR does
~~Swaps the map in space for one stationed on the ground. There are two shuttles **(one with 33 seats and the ability to load from cargo, another with three mining drills and and outlet to the smelter on the station.)** **32 drop pods**, and a gateway if the FOB is being put on the next map. This can easily be turned into a space themed map if need be, I just like the jungle aesthetic. OLD MAPS HAVE BEEN BACKED UP- I think. _core\oldstation2, backup\station_~~

Adds a new 64x prefab- the syndicate mining outpost, using the bits I liked from my cobbled together map.

map

Something something prefab, something something.


Moltovz:
should maints be bigger, yes
do we need two shuttles, I don't know, I haven't seen the concept for the full new map yet.

BurgerLUA (approved):

PR #468, opened 2022-03-09
Throwing dragged

Chip11-n:
# What this PR does
Allow you to throw dragged items and mobs(only with agressive grab).
Currently has only visual bug that causes you to "hit" target like in melee, but as I know, fix is on the way.
As another thing, if something throwed has "/damagetype/error", overrides it as stunbaton's /damagetype/

Old-new feature, Trello

### DIFF
--- a/code/_core/obj/hud/inventory/interaction.dm
+++ b/code/_core/obj/hud/inventory/interaction.dm
@@ -43,29 +43,46 @@
 		if(!isturf(caller.loc))
 			return TRUE
 		var/mob/living/L = caller
-		caller.face_atom(object)
-		var/atom/movable/object_to_throw = top_object
+		caller.face_atom(object) //Changing dir
+		var/atom/movable/object_to_throw
+		if(grabbed_object) object_to_throw = grabbed_object //We want to throw something that we pull
+		else object_to_throw = top_object //Else we throw something in our hand
+
 		if(istype(object_to_throw))
-			var/obj/item/I = object_to_throw
-			if(I.additional_clothing_parent || I.no_drop)
-				caller.to_chat(span("warning","You can't throw this!"))
-				return TRUE
-			var/vel_x = object.x - caller.x
+			if(ismob(object_to_throw) && grab_level < 2) //To throw mob you need agressive grab
+				caller.to_chat(span("warning","You need a better grip to do that!"))
+				return TRUE //Isn't agressive grab on mob?
+
+			var/vel_x = object.x - caller.x //Caller pos is the same as object anyway
 			var/vel_y = object.y - caller.y
 			var/highest = max(abs(vel_x),abs(vel_y))
 
-			if(!highest)
-				I.drop_item(get_turf(caller)) //Drop if we can't throw.
+			if(highest == 0) //Our position is the same as target
+				release_object(caller) //Better not to divide by zero
 				return TRUE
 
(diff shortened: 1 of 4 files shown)


code/_core/obj/hud/inventory/interaction.dm
```diff
@@ -43,29 +43,46 @@
...
+				caller.to_chat(span("warning","You need a better grip to do that!"))
+				return TRUE //Isn't agressive grab on mob?
+
+			var/vel_x = object.x - caller.x //Caller pos is the same as object anyway
 			var/vel_y = object.y - caller.y
 			var/highest = max(abs(vel_x),abs(vel_y))
 
-			if(!highest)
-				I.drop_item(get_turf(caller)) //Drop if we can't throw.
+			if(highest == 0) //Our position is the same as target
```
BurgerLUA, line 60: This should be below the item checks because I.drop_item() won't get called on line 70 ever.

code/_core/obj/hud/inventory/interaction.dm
```diff
@@ -94,8 +111,21 @@
 			drop_item_from_inventory(get_turf(src))
 		return TRUE
 
-	if(grabbed_object && grabbed_object == object)
-		release_object(caller)
+	if(grabbed_object && grabbed_object == object && is_living(grabbed_object)) //We click on the hand that grabbed something
+		if(world.time <= grab_time+DECISECONDS_TO_TICKS(20)) //Prevents insta agressive-grab
+			caller.to_chat(span("warning","Too soon to tighting your grip on [object]! Need to wait another [TICKS_TO_SECONDS(grab_time+DECISECONDS_TO_TICKS(20) - world.time)] seconds!"))
```
BurgerLUA, line 116: A to_chat shouldn't be here. On 99% of features, we don't ever tell the player if they did something too soon.

code/_core/obj/hud/inventory/interaction.dm
```diff
@@ -94,8 +111,21 @@
...
+			return TRUE
+		caller.to_chat(span("warning","You tighten your grip on [object]!"))
+		grab_level = 2 //Agressive grab
+		grab_time = world.time //We grabbed now. Used in agressive grab
+
+		var/mob/living/A = owner //How can something not alive grab anything?..
+		A.add_status_effect(SLOW,30,30) //I don't know any better way..
+		var/mob/living/V = grabbed_object //Checked above
+		V.add_status_effect(REST,30,30)
+		V.add_status_effect(SLOW,30,30)
```
BurgerLUA, line 126: Just to let you know, all of theses status effects will only last for 3 seconds. Is that intended?

BurgerLUA (changes requested):
Looks good but there are a few issues.

code/_core/obj/hud/inventory/interaction.dm
Chip11-n: Initially, I wanted to make it permanent (or until drop), but then looking at the experience from other servers, I thought that permastun would be far from the most interesting feature. If you managed to grab something, you have upper hand for some time, but it shouldn't be the end.

code/_core/obj/hud/inventory/interaction.dm
Chip11-n: Done.

BurgerLUA (approved):

PR #469, opened 2022-03-10
Changes some vendors and their palettes

SmArtKar:
# What this PR does
Just what the title says.

Old sprite example using a mushy gradient pallet
image

New sprite example using existing color palettes
image

Consistent color palettes and less gradient mush are always good.


BurgerLUA (approved):
nice

PR #474, opened 2022-04-14
Fixes gauss inhands

Moltovz:
# What this PR does
Says it on the tin

flotaing gunz bad!!11!1


Stalkeros2 (approved):
yes

PR #478, opened 2022-05-10
Chargen on station removal + The range

Moltovz:
# What this PR does
Adds a chill zone for THE BOYS to Bluespace.DMM.......
**also replaces the chargen markers on the station's bathroom with backup spawners because there shouldnt be chargen ones on the station anyway.....**
image

If you can add the backrooms, I can add a place to hang out with the BOYS and barbeque..... Until we actually need the space in bluespace.


BurgerLUA:
Is this basically a sandbox area?

Moltovz:
> Is this basically a sandbox area?
literally

Moltovz:
Closed because Bluespace can be utilized better. Charge.

PR #480, opened 2022-07-03
You can wield the tungsten SMG now.

Moltovz:
# What this PR does
Title. Bugfix. Shooting the tungsten SMG one-handed will cause you to cry in pain and drop it. It's supposed to be wieldable.

bugfix.


Stalkeros2 (approved):
I guess?

PR #481, opened 2022-07-12
Holy m1911 fix? seems to work.

NzzyFk:
# What this PR does
Makes the Holy M1911 work.

So the funny holy stun gun works!


Stalkeros2 (approved):
I agree.

PR #510, opened 2022-07-27
Medipen price nerf

Stalkeros2:
# What this PR does
Reduces mining butterfly pen's price from 2500 to 1000, effectively making them actually worth buying. Also replaced synthflesh with tricord to make it non- (or less-) toxic.

You can buy roughly two medkits for a price of one mining pen, thus making it almost entirely useless for medicinal purposes. This should fix that.
Chem replacement is me making sure that nobody dies while using one.


Stalkeros2:
After doing some testing, I have come to a conclusion that you literally become immortal for brief seconds after using one, doubling that time if injected with two or more. Even then, it'd still be a good idea to reduce its price, because the user literally drops dead afterwards.

BurgerLUA (approved):

PR #521, opened 2022-09-03
Legion core stabilizer

Stalkeros2:
# What this PR does
Adds legion core stabilizers that're capable of preserving legion cores, making them have no expiry time.
Buffs legion cores to heal most common damage types and then delete themselves upon use.
Legion cores now use themselves on the user (duh) when used in hand.
This'll make them way more useful than they're currently.


BurgerLUA (approved):
Good but I'm going to have to nerf this in some way.

PR #530, opened 2022-09-05
Gold Toolbox value down

ChewedPencil:
Lowers the value so it ACTUALLY spawns in trash piles

# What this PR does
WHat it says on the tin baybeeeee

Cause my beloved golden box was IMPOSSIBLE TO FIND


Moltovz:
I get the blunt AP change, though it is a little bit steep. What's with the fifth of Holy AP? Also iirc value = doesn't actually matter when selling something. @BurgerLUA correct me if I. Wrong.

ChewedPencil:
The lowering of the cost was to make it available to spawn in trash loot pool, as 5000 credits is the maximum an item can be worth and still show up. I noticed it while digging around the loot datjm for the trash piles.

As for the holy ap, i feel like i had originally added it cause i assumed that it would make more holy damage come trough armor. But im starting to think it  is fo4 ap for that specific armor type and not all armor

BurgerLUA (approved):

BurgerLUA:
> I get the blunt AP change, though it is a little bit steep. What's with the fifth of Holy AP? Also iirc value = doesn't actually matter when selling something. @BurgerLUA correct me if I. Wrong.

I just realized that it only considers the raw initial value and not the calculated value so I might have to overhaul how trash spawning works.

PR #556, opened 2022-10-24
Deletes exactly 2 bug?(s)

V2LenKagamine:
# What this PR does
Causes spellswapping a spellswapped item to Qdel the old spellswap instead of the item itself
Makes staffs remember how many slots they have on save/load

Not funni bug bad, funni bug acceptable.

These two not funni.


Stalkeros2 (approved):
hghkglghlhggyirydqwauiihu75e543235

BurgerLUA:
figured this was the issue lmao

PR #557, opened 2022-10-24
Enchanting, but moff & lizorb

V2LenKagamine:
# What this PR does
Allows moff to enchant
Yes I actually tested it, yes it works.

Also removes all " ' " from enchanting phrases because seriously they just cause issues with my moth code.

The idea of a moff making a flame XXXI torch appeases me.


V2LenKagamine:
Crap left in a debug line, dont merge yet ill fix in morning

Stalkeros2 (approved):
fair enough

PR #563, opened 2022-10-25
[Bugfix] Rounding-B-Fix

V2LenKagamine:
# What this PR does
Fixes that god forsaken ROUNDING BUG that causes containers to start with 0.001 less and be unable to be filled fully.
The fix isn't exactly perfect, you can still occasionally end up with 0.001 of a reagent left in very specific scenarios, but it *shouldn't* cause any 1st law of thermodynamics breakage.

Hmm yes today I would like to have every container permanently unfillable and act like a black hole when almost full.

But seriously this bug sucked. My solution probably isn't perfect but its *better than the bug*


Stalkeros2:
This looks like an actual code so I can't approve it as I can't code.
Will have to wait for Burger ig.

BurgerLUA:
I'll have to merge it and then take a closer look at it. I think the round() proc should still be used in the add_reagent proc because some fuckery might occur with transfers.

V2LenKagamine:
Ill do some more testing and see if any odd situations come up

V2LenKagamine:
So after some *arguably minimal* testing, I note a few odd cases:
1: When making a new reagent,(EG:Dough) it is occasionally possible to end up with .001 of a reagent not being converted in the recipe, leaving .001 of the product unmade, avoided by mixing in "perfect" quantities. Most likely a recipe portion error, and the bigger of the issues I found
2: If the components can not be *perfectly* divided by the transfer amount(EG: 10 of 3 differnt materials, transfer 10), it will transfer +-.001 to ensure that *no material is voided*. This leaves an odd message in chat saying it transferred 10.001/9.999, but at least no material is voided. Not a massive deal in my opinion but Im weird.

V2LenKagamine:
Burmger made better fix, no longer required.

PR #569, opened 2022-10-26
Disables meatmen squads from spawning on unclean chunks.

Moltovz:
# What this PR does
says it on the tin

Speedmerge this shit, server performance.


BurgerLUA:
I nerfed beefmen so they should be weaker now.

PR #583, opened 2022-10-27
Skat-9 no longer makes feet invisible.

Moltovz:
# What this PR does
Removes "left and right foot" protection from the Skat-9 and skat-9 antag. They don't have visible armor? They don't GET the armor. This also stops feet from going invisble, so thats cute.

bugfix


BurgerLUA:
Does this though? There should be a hidden limbs list that does what you think.

BurgerLUA:
closing to fix this myself

PR #590, opened 2022-11-02
[Bugfix]Bugz b gone

V2LenKagamine:
# What this PR does

Fix sus inability to put in items
Fix artifacts being fultonable
Fix plasma charger being charged by anything
Fix negative med XP
Fix deathbox not giving all items
Reduces sound of fire effect by half because LOUD
Fix ability to turn in contracts at any point.
Makes drop-pods less jank by self-deleting after 3 0 seconds but allowing old landing conditons.

Bug bad


code/_core/obj/structure/interactive/drop_pod.dm
```diff
@@ -116,18 +116,17 @@ var/global/list/turf/drop_pod_turfs = list() //Drop pods that need to respawn.
...
+			if(T.has_dense_atom)
+				caller.to_chat(span("warning","Invalid drop location: Target area is obstructed."))
+				return FALSE
+			if(!T.is_safe() || !T.is_safe_move())
+				caller.to_chat(span("warning","Invalid drop location: Unsafe area."))
+				return FALSE
+			var/area/A = T.loc
+			if(A.interior)
+				caller.to_chat(span("warning","Invalid drop location: Target area is obstructed."))
+				return FALSE
```
BurgerLUA, line 129: should probably revert the changes here as this is old drop pod turf checking code

code/_core/obj/structure/interactive/drop_pod.dm
```diff
@@ -154,7 +153,9 @@ var/global/list/turf/drop_pod_turfs = list() //Drop pods that need to respawn.
 			for(var/k in contents)
 				var/atom/movable/M = k
 				M.force_move(src.loc)
-
+			CALLBACK("set_state_\ref[src]",50,src,.proc/set_state,caller,POD_DESTROY,desired_loc)
+		if(POD_DESTROY)
+			src.Destroy()
```
BurgerLUA, line 158: qdel should be used here instead of src.Destroy()

BurgerLUA (changes requested):

code/_core/datum/reagent/reagent_medicine.dm
```diff
@@ -15,7 +15,7 @@
 	. = ..()
 
 	if(!L.dead && experience_per_unit*. > 0 && is_player(caller) && caller.client)
-		caller.add_skill_xp(SKILL_MEDICINE,experience_per_unit*.)
+		caller.add_skill_xp(SKILL_MEDICINE,CEILING(experience_per_unit*.,1))
```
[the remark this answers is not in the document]
V2LenKagamine: This actually wasnt the issue, what we found out was there was a missing negative sign, which I due believe you caught and already fixed, but this also ensures that you always gain at least one XP for a reagent so.

code/_core/obj/item/contract/_contract.dm
```diff
@@ -107,18 +107,22 @@
 
 /obj/item/contract/proc/turn_in(var/mob/living/advanced/player/P,var/params)
 	var/turf/T = get_turf(P)
-	if(burgerbux_reward)
-		P.to_chat(span("notice","You are awarded \the [reward.name] and [burgerbux_reward] burgerbux for completing the contract."))
-		P.adjust_burgerbux(burgerbux_reward)
+	if(src.value_current != src.value_max)
```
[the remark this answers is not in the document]
V2LenKagamine: Point taken.

code/_core/obj/structure/interactive/drop_pod.dm
V2LenKagamine: This is actually preferred from my experience, as many players would come in and ask if "drop pods were bugged" as it would often seemingly return false positives. The new delete code should fix the issue of it blocking off key areas.

code/_core/obj/structure/interactive/drop_pod.dm
V2LenKagamine: Wouldnt calling Qdel leave it in the subsystem for all drop pods? Calling destroy here makes sure to remove it from that subsystem by properly running its destroy code.

BurgerLUA: Nope. qdel runs src.Destroy properly. qdel is basically calling Destroy but with some extra checks

code/_core/datum/reagent/reagent_medicine.dm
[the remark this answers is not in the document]
BurgerLUA: understandable

code/_core/obj/item/contract/_contract.dm
```diff
@@ -107,18 +107,22 @@
 
 /obj/item/contract/proc/turn_in(var/mob/living/advanced/player/P,var/params)
 	var/turf/T = get_turf(P)
-	if(burgerbux_reward)
-		P.to_chat(span("notice","You are awarded \the [reward.name] and [burgerbux_reward] burgerbux for completing the contract."))
-		P.adjust_burgerbux(burgerbux_reward)
+	if(src.value_current <= src.value_max)
+		P.to_chat(span("warning","You feel it would be unwise to try to turn in a contract that isn't complete yet!"))
+		return FALSE
```
BurgerLUA, line 112: I made a mistake, which is my bad. It should actually be < instead.

V2LenKagamine: And I made the mistake of being to tired to logic check.

BurgerLUA (approved):

PR #592, opened 2022-11-06
Lays out the foundation of the Voice Of God Ability Package

FancyCyclops:
# What this PR does
Open's pandora's box. 

Adds a bunch of abilities that (at the very least) cause status effects to enemies that enhance the damage dealt to them, as well as providing an ability that heals allies (and staves off their deaths).

Prayer will have influence over the power of this ability set.

More ability variety is nice. More customizability is nice.

# To do:
-Find a non-jank way to call the damage check
-Figure out what status effect Bleed should do
-Implement a status for Harm
-Figure out how laggy this will end up being
-Figure out how to add an exemption for abilities in the same catagory
-Figure out how to avoid the credit.

### DIFF
--- a/code/_core/datum/ability/voice_of_god.dm
+++ b/code/_core/datum/ability/voice_of_god.dm
@@ -1,0 +1,73 @@
+/ability/voice_of_god/harm
+	name = "Command: Harm"
+	desc = "Damages all enemies within 8 tiles, and makes them susceptable to further damage."
+	icon_state = "dash"
+
+	resource_type = MANA
+	var/damagetype/damage_type
+	var/atom/caller
+	var/word = "DIE"
+	category = "Voice Of God"
+
+/ability/voice_of_god/harm/proc/get_params(var/atom/victim)
+	. = list()
+	.[PARAM_ICON_X] = rand(0,32)
+	.[PARAM_ICON_Y] = rand(0,32)
+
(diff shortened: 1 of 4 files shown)


code/_core/datum/ability/voice_of_god.dm
```diff
@@ -0,0 +1,73 @@
...
+	var/mob/living/L = caller
+	var/turf/T = get_turf(L)
+	play_sound('sound/effects/invoke_general.ogg',T, range_max = SOUND_RANGE)
+	L.do_say("<font color='#DD1C1F' size='4'>[word]</font>",FALSE)
+
+	for(var/mob/living/L2 in oview(T,8))
+		if(L2.dead)
+			continue
+		if(L2.loyalty_tag == L.loyalty_tag)
+			continue
```
BurgerLUA, line 42: should replace this with !allow_hostile_action(L2.loyalty_tag,L)) and remove the one on line 22.

FancyCyclops: Done and done. Should look better now.

FancyCyclops:
Attempting to fix this up has become an increasingly frustrating endeavor. So that people can be get the motivation to pick up my slack if they decide to work on this, I've decided to close this off until I either rattle something in my head up right or this collects enough cobwebs to be a spider's nest.

PR #593, opened 2022-11-07
[Content][Bugfix]Burger-craft

V2LenKagamine:
# What this PR does
Re-impliments crafting as well as several new items related to crafting/gathering ingredients including
-Mining
-Magical Botany
Allows vendors to specify a per-item multiplier to item cost including the per-vendor multiplier to cost.
Re-impliments alchemy, the magical, non-sensical version of chemistry. Recpies for this need not reagents, just items.
burmger crafmt. Whats a game without crrafting anyway?


Stalkeros2:
Freaking madman.

V2LenKagamine:
Ok, the framework for crafting has been set. Anyone can now easily add recipes for items with a simple format.
Recipes are able to say:
How much of each item you get per craft. (Yes you can get multiple items from one craft)
How much is taken from each slot
What must go where
What can be anywhere
Chance to succeed on the craft per item
What you get on failing
What reagent is put in the container after crafting
All in all, a robust system, but it has its glitches and quirks. You cannot craft with anything that has a contents list greater than 0.
The GUI is buggy, which I am unable to fix, and for some reason ammo crafting more than 1 of a mag divides by 0? No idea. Its out of my skill range. Currently its WORKABLE however.
This also adds some (Poorly sprited) base crafting items such as gunpowder and shells.
Now including random ass vending machine changes and some DU ammo for specific weapons!

code/_core/mob/living/advanced/human/npc/beefman.dm
```diff
@@ -55,4 +55,11 @@
...
-	return PLANE_MOB_STEALTH //Always stealth.
\ No newline at end of file
+	return PLANE_MOB_STEALTH //Always stealth.
+/mob/living/advanced/npc/beefman/on_walk()
+	var/turf/T = get_turf(src)
+	for(var/obj/effect/cleanable/B in T.contents)
+		qdel(B)
+		src.health.health_max += 5
+		src.health.adjust_loss_smart(-5,-5,-5)
+	. = ..()
```
BurgerLUA, line 65: I don't think this should be added. This will be super laggy.

BurgerLUA (approved):
lot of stuff here that's wrong but I guess I can fix it myself

PR #595, opened 2022-11-07
Reagent Metabolism Overhaul + Syringe Bugfixes

BurgerLUA:
Overhauls reagent metabolism so it is more balanced and less prone to funny exploits. Also fixes a few bugs.

## Feature Changes
- Metabolism limits are introduced. You can only metabolize up to a certain volume of reagents at a time. Prevents weird exploits like 50 different types of weaker poisons being more effective than cyanide because of it being all metabolized at once.
- If you have less than 1u of a certain type of reagent in your bloodstream, that reagent is NOT metabolized and slowly removed from your system. This also prevents weird exploits.
- If you have more than 1u of a certain type of reagent in your bloodstream, but metabolizing it would bring it under 1u, all of that reagent is metabolized at once. This makes microdosing (from burgers, smokes) much more faster.
- Blood toxicity is properly displayed in the medical analyzer.

## Bug Fixes
- Fixes syringes not injecting properly.
- Fixes some medicines that use on_add_living and on_remove_living from not being properly checked.
- Fixes syringes and hyposprays having the wrong type path.


ilikesaying:
how does this affect dionae, who literally consume 100u of reagents

BurgerLUA:
> how does this affect dionae, who literally consume 100u of reagents

They'll metabolize it slowly.

PR #606, opened 2023-03-16
[wip] market

jjpark-kb:
# What this PR does
Adds the market-- a place where players can choose their price to sell an item; then, another player can buy that item for that sold price.

Safe ways to guarantee a player can sell an item and allow other players to actually purchase the item (rather than it being voided).


jjpark-kb:
It somewhat works-- still not merge-able state:
1) you can sell the item for an amount (and get the credits for selling it)
2) you can see the item in the inventory to buy
3) you can spend your money to get the item, but you don't get it

jjpark-kb:
closing for now.

PR #611, opened 2023-05-13
Botal-alchemy Overhaul. Alchemy now is proper plant.

V2LenKagamine:
# What this PR does
All Alchemy plants have been turned into *proper* plants. Seeds and all.
Currently, the only way to obtain them would be getting lucky in the wild, or using a seed from another plant.
Disables old alchemy_plant system and related.
Adds a new reagent: Armo-tol! Provides stoneskin buff while being metabolized.

Old alchemy plant system was, to be honest, completely hacked together to be a more performant botany system for "scavenged" or ungrowable plants. This has its pros and cons, one notable con being that they couldnt use reagents. This new system would theoretically be more heavy than the old one, but much more flexible and probably better coded.

# Notes
Sprites are still very much coder tier sprites. Someone please make better ones.
Compiles, and *should* work, but only one plant was fully tested, meaning I may have missed a sprite.


code/_core/datum/reagent/reagent_magic.dm
```diff
@@ -24,4 +24,59 @@
 		L.add_stand()
 		if(current_volume) container.remove_reagent(src.type,current_volume,should_update = FALSE, check_recipes = FALSE)
 		return 0
+*/
+
+/reagent/magic/armotol
```
BurgerLUA, line 29: This should be a subtype of /reagent/drug/ as it already has a thing to handle status effects. Check out the file reagent_drug.dm for how it works.

BurgerLUA (changes requested):

code/_core/datum/reagent/reagent_magic.dm
V2LenKagamine: Completed, thanks for showing me that.

BurgerLUA (approved):

PR #613, opened 2023-05-15
Make it Burn-dem: Incen explosives and Flamethrower from a distant era...

V2LenKagamine:
# What this PR does
Adds 40mm Incen grenade
WP rocket now creates a small fire, but high damage fire as well as the explosion.
Adds a flamethrower from a distant era, feature complete, with ad-meme variant.
# Notable code changes

Weapons 'shoot()' can now be given the 'projectile_override' variable to call shoot from the weapon while forcing a different projectile than the default, presuming it fails, uses default projectile. Takes the path of projectile as arguement, default null

New weapon Type: Reagent_ammo. For things that consume a reagent as ammo.

New Object: Reflector. Reflects things back from whence they came.

Code Change: Projectiles call the on_projectile_hit of what they slam into. So far, no major side effects.
References are good. Also, even if you dont like the flamethrower, free code for other reagent as ammo weapons.
# Notes
reflector could use a better sprite and animation, but not my forte sadly.
Does not add a way to obtain either flamethrower. Someone suggested new pyro unique mob?


V2LenKagamine:
Alright, after running a few more tests im confident in saying that my changes had no major side effects to projectiles, and things still run smoothly. Found a couple bugs in testing but should have patched them out. Ready for Review.

code/_core/obj/item/weapon/ranged/_ranged.dm
```diff
@@ -386,15 +386,15 @@ obj/item/weapon/ranged/proc/play_shoot_sounds(var/mob/caller,var/list/shoot_soun
...
+obj/item/weapon/ranged/proc/shoot(var/mob/caller,var/atom/object,location,params,var/damage_multiplier=1,var/click_called=FALSE,var/projectile_override = null)
 
 	if(!pre_shoot(caller,object,location,params,damage_multiplier))
 		return FALSE
 
 	var/quality_bonus = get_quality_bonus(0.5,2)
 	var/quality_penalty = max(1,1/get_quality_bonus(0.25,2))
 
-	var/obj/projectile/projectile_to_use = projectile
+	var/obj/projectile/projectile_to_use = (projectile_override != null && ispath(projectile_override, /obj/projectile) ? projectile_override : projectile)
```
BurgerLUA, line 397: I would probably remove the ispath check here because, believe it or not, ispath and istype checks are expensive. If it causes errors because of a bad projectile_override var, then that's on the coder to fix.

code/_core/obj/item/weapon/ranged/unique/_unique.dm
```diff
@@ -0,0 +1,20 @@
+/obj/item/weapon/ranged/unique/
```
BurgerLUA, line 1: Not sure I like this tbh. I rather you make a subtype for a weapon that takes reagents as ammo instead of a catch-all type that takes anything.

code/_core/obj/item/weapon/ranged/unique/flamethrower.dm
```diff
@@ -0,0 +1,87 @@
...
+/obj/item/weapon/ranged/unique/flamethrower/get_ammo_count()
+	return FLOOR((reagents.volume_current/fuel_per_shot),1)
+
+/obj/item/weapon/ranged/unique/flamethrower/can_gun_shoot(mob/caller, atom/object, location, params, check_time, messages)
+	if(get_ammo_count() <= 0)
+		handle_empty(caller)
+		return FALSE
+	return ..()
+
+/obj/item/weapon/ranged/unique/flamethrower/click_on_object(mob/caller, atom/object, location, control, params)
```
BurgerLUA, line 42: I'd put this in shoot() instead of click_on_object.

code/_core/obj/projectile/_projectile.dm
```diff
@@ -203,6 +204,16 @@
...
+	for(var/obj/projectile/other_P in new_loc.contents) //Handles P on P collisions
+		if(projectile_should_collide(other_P,old_loc,new_loc) || other_P.projectile_should_collide(src,old_loc,new_loc))
+			if(other_P.on_projectile_hit(src,old_loc,new_loc)) //Checks if the other projectile should do something to this one.
+				other_P.penetrations_left--
+				//Dont return anything here, it may have more to its life, but note it did hit something.
+			if(on_projectile_hit(other_P,old_loc,new_loc)) //Checks if this projectile should do something to the other one.
+				penetrations_left--
+				if(penetrations_left < 0)
+					return FALSE //This projectile has served its purpose.
+
```
BurgerLUA, line 216: This code unfortunately will not work as intended because of how projectiles work.

When a projectile is created, it does not actually move. It stays in one location and the movement that it has is completely simulated to bypass byond's movement to reduce the amount of lag projectiles have.

The only feasible way to add this feature is to make it so that the pyro airblast creates a special type of /obj/ that projectiles can collide with via the projectile_should_collide() proc so that moving projectiles can detect this /obj/.

code/_core/obj/projectile/laser.dm
```diff
@@ -93,11 +93,7 @@
...
-
+/obj/projectile/bullet/flamer/on_enter_tile(turf/old_loc, turf/new_loc)
 	. = ..()
-
-	var/obj/effect/temp/hazard/flamethrowerfire = locate() in new_loc
-
-	if(!flamethrowerfire)
-		new /obj/effect/temp/hazard/flamethrowerfire(new_loc,SECONDS_TO_DECISECONDS(30),owner) //time and hazard type not final
\ No newline at end of file
+	if(new_loc != get_turf(owner)) // this stops you setting fire to yourself, unless you walk into the flames...
```
BurgerLUA, line 98: I feel like there is a better way to fix this. If flamethrower fire is getting created on the turf that the owner is on, then there is something wrong with something else.

BurgerLUA (changes requested):

code/_core/obj/item/weapon/ranged/_ranged.dm
V2LenKagamine: Noted for future reference, will remove.

code/_core/obj/item/weapon/ranged/unique/_unique.dm
V2LenKagamine: Probably just me overthinking weapon ideas and preparing for them, I'm good with unique weapons, but for now, poof. It shall disappear.

code/_core/obj/item/weapon/ranged/unique/flamethrower.dm
V2LenKagamine: Yea. Point.

code/_core/obj/projectile/_projectile.dm
V2LenKagamine: Believe it or not, not only does this code work, it works as I intended it to!
Because of it being coded in an either-or sense, it doesnt care if a projectile enters its tile or if it enters a tile with a projectile, the airblast detects it and reflects properly. This also allows anyone to make their projectile, say, absorb other projectiles to make itself stronger. I've tested this for a few rounds, and it works superbly; I've noticed no major performance degradation from this change.

Though if you're not satisfied with this solution, let me know and I will gladly find another solution.

code/_core/obj/projectile/laser.dm
V2LenKagamine: I'll give a poke around, but when the projectile spawns, it instantly sets fire to the tile it spawned on, and because shoot spawns bullets on YOUR tile, Fwoosh. I don't know if it is intended to trigger on_enter_tile on spawn, but my guess is it does.

code/_core/obj/item/weapon/ranged/unique/flamethrower.dm
V2LenKagamine: Turns out I actually had it there because otherwise you attempt to spill contents instead of airblasting.

code/_core/obj/projectile/_projectile.dm
BurgerLUA: This might create issues where you can airblast bullets where they spawned and those bullets will be reflected even though it shouldn't be. I highly recommend doing my solution.

V2LenKagamine: Took me a moment but I get you now, will change solution

BurgerLUA (approved):

PR #626, opened 2023-06-15
SKAT-10 now should treat its user's damage

Stalkeros2:
# What this PR does
Wearing SKAT-10 now makes its user regenerate brute, burn, pain damages, as well as blood, as intended.

Makes this particular super rare armor a bit more unique; and actually working the way it's meant to, turning its user into a light-to-medium damage sponge.


BurgerLUA:
I'm a little conflicted about adding equipment that gives self-regen, even if it's small.

PR #634, opened 2023-06-23
Nerfs the Bluespace Satchel to give Bluespace Backpacks more purpose

genessee-forgot-his-password-again:
# What this PR does

This PR makes the Satchel of Holding able to hold 28 items (7x4 grid) compared to the Bag of Holding's 32 items (8x4 grid).

Additionally, this PR increases the value of Bags of Holding up from 3000 to 5000, and gives them the 'Rare' rarity.
Additionally, this PR makes Satchels of Holding have the 'Uncommon' rarity.
Additionally, this PR makes Satchels of Hold, the prank item, have the 'Common' rarity and decreases their value to 500 to match their prank item status while still holding a little bit of value for use in selling to the shop.

Right now, it feels to me like Bags of Holding are just a worse Satchel of Holding in nearly every way, as Satchels have the same storage space (except being unable to carry heavy guns, I think?) while . I got the impression that Satchels of Holding were supposed to at least have less storage slots than the Bag of Holding to give more reason to actually use Bags of Holding over Satchels of Holding.

Satchels of Holding are still very large and can hold more items than rudimentary dufflebags (which have 8x3, or 24 slots), as they themselves are also treasure items, but still have less storage capacity than BoH's along with the inability to store large items.

As Bags of Holding are the largest storage item in the game, I found it odd they were considered 'Common' - probably an oversight, but y'know. Fixed that.

### DIFF
--- a/code/_core/obj/item/clothing/back/storage/bluespace/_bluespace.dm
+++ b/code/_core/obj/item/clothing/back/storage/bluespace/_bluespace.dm
@@ -18,9 +18,11 @@
 		/obj/item/clothing/belt/holding
 	)
 
-	value = 3000
+	value = 5000
 
-	value_burgerbux = 1
+	value_burgerbux = 2
+
+	rarity = RARITY_RARE
 
 /obj/item/clothing/back/storage/satchel/bluespace
 	name = "satchel of holding"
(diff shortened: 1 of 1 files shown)


BurgerLUA:
did some changes myself taking inspiration from this pr

my nerfs are heavier though

PR #639, opened 2023-06-28
Adds OSRS hit sound to Runic Scimitar

muzcl:
https://github.com/BurgerLUA/burgerstation/assets/109846546/06359426-ab45-4e56-af64-39a297e99f07


muzcl:
unmute audio btw

Luxzhv:
I'd personally like the audible "TING" sound rather then the generic painhurt sound, but it's iconic enough.

muzcl:
> I'd personally like the audible "TING" sound rather then the generic painhurt sound, but it's iconic enough.

Do you mean the other sound that plays when I attack in the video? I don't think I can get rid of that without fucking with the code, and that's more than I'm capable of

BurgerLUA (approved):

PR #648, opened 2023-07-14
Allow the 20mm AMR to take barrel chargers

irRegularGuy646:
# What this PR does
adds the barrel charger and advanced barrel charger to the 20mm sniper's attachment whitelist

overkill? more like, just enough kill


BurgerLUA:
Wouldn't work because the barrel of the gun itself is an attachment (for a larger sprite) so it would look funny.

Also I am not in favor of adding barrel attachments for extremely powerful weapons.

irRegularGuy646:
just discovered that the corrupting frog is a thing

PR #657, opened 2023-07-17
Adds Fairy Wings

muzcl:
# What this PR does
Adds Fairy Wings item, sprites are from an unused fairy_wings.dmi in the files. It gives +40% speed, +15% more than other wings, but it gives you -100 armor against every type but arcane, which is +50.

I didn't want to leave the sprite there to rot, and having a different wing choice is nice. Unfortunately, I couldn't get it to be dyable without it fucking up how its supposed to look normally.


muzcl:
they're only obtainable by corrupting frog

BurgerLUA (dismissed):

muzcl:
New commit added: Fairy enemy, drops Fairy Wings, it spawns from anomalies with the same rarity as Dick (the new wings are better in speed, so we need rarity to compensate), and Fairy Wings are now dyable

BurgerLUA (approved):

PR #666, opened 2023-08-19
.45 LongCon invisible sprite fix

Ctr-Truescreen:
# What this PR does
Title. Fixes a typo in the .dmi file, which was the root cause of the problem.

bugfix good :)

image


TypeVar:
finally, can use the syndicate 45. Where do you even get it?

Ctr-Truescreen:
Syndicate armament vendors are located either inside the vault area of Traitor Joe's inside the Rael city area, or inside the syndicate ship south-west of Landing Zone 9. The syndie ship is a guaranteed spawn, but the location of Traitor Joe's may vary slightly.

BurgerLUA (approved):

PR #671, opened 2024-01-24
Ports time.dm defines from TGStation

Xander3359:
# What this PR does
Ports time.dm defines from TG
deletes MINUTES_TO_SECONDS(x) It was unused, so nothing worth noting.
Replaces all instances of SECONDS_TO_DECISECONDS with the new macro SECONDS
Replaces TICK_LAG with TICKS
Replaces DECISECONDS_TO_TICKS with DS2TICKS
Replaces TICKS_TO_DECISECONDS with TICKS2DS

Better code, very cool.


BurgerLUA:
I'm conflicted about this PR since there is a lot of things in the new file that are unused, and I vastly prefer my system to /tg/'s system.

BurgerLUA (approved):
Will merge tomorrow when I wake up just in case there are unforeseen issues.

PR #672, opened 2024-01-24
Updates math.dm to what TG has

Xander3359:
# What this PR does
Updates math.dm to whatever TG has
It is now named maths.dm (to match TG's filename)

Updated code, stay modern coders 😎

### DIFF
--- a/burgerstation.dme
+++ b/burgerstation.dme
@@ -68,7 +68,7 @@
 #include "code\__defines\lighting.dm"
 #include "code\__defines\loot.dm"
 #include "code\__defines\material.dm"
-#include "code\__defines\math.dm"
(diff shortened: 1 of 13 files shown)


BurgerLUA (changes requested):
since the file was renamed from math to maths, it is no longer included in the .dme

Xander3359:
> since the file was renamed from math to maths, it is no longer included in the .dme

I addressed this issue in the latest commit.
I initially was very confused because I couldn't find the .dme file somehow? 
But that was an error on my part since I found it now.

BurgerLUA (approved):

LemonInTheDark:
This pr breaks our license

PR #675, opened 2024-01-26
Prestige now sets you to Prestige Level x 5 instead of just 5

Xander3359:
# What this PR does
When you prestige, your level will be set to 
Prestige level * 5
instead of just
5

Adding a baseline level makes it so that future prestiges don't feel shitty.
It takes a while to grind up to level 100 (and above)
It feels terrible that after all that time spent levelling, you don't really *gain* anything until you start hitting those level 101+

Giving it a baseline of prestige * 5 makes you feel like you are actually progressing.
And it's not too unreasonable balance-wise since those early levels tend to be the easiest to get (due to low exp reqs)

### DIFF
--- a/code/_core/mob/living/advanced/player/_player.dm
+++ b/code/_core/mob/living/advanced/player/_player.dm
@@ -1,6 +1,7 @@
-/mob/living/advanced/player/
+/mob/living/advanced/player
 
-	var/unique_pid //Snowflake system that generates a md5 hash of the player on character creation.
+	///Snowflake system that generates a md5 hash of the player on character creation.
+	var/unique_pid
 
 	desc = "Seems a little smarter than most, you think."
 	desc_extended = "This is a player."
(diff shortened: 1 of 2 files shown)


code/_core/mob/living/advanced/player/_player.dm
```diff
@@ -221,12 +229,12 @@
...
 	return TRUE
 
-/mob/living/advanced/player/proc/prestige(var/skill_id)
+/mob/living/advanced/player/proc/prestige(skill_id)
 	if(!prestige_count[skill_id])
 		prestige_count[skill_id] = 1
 	else
 		prestige_count[skill_id] += 1
-	set_skill_level(skill_id,5)
+	set_skill_level(skill_id, (prestige_count[skill_id]*5))
```
BurgerLUA, line 237: I'm iffy about this PR because theoretically if you prestige 20 times you can then prestige an unlimited amount of times.

BurgerLUA: Actually nevermind, I understand.

BurgerLUA (approved):

PR #680, opened 2024-01-31
makes doctor bag actually pay the doctor

Xander3359:
# What this PR does
Doctor bag now actually pays the doctor for healing

Makes the thing work like it's supposed to.


BurgerLUA:
failing checks because of code/_core/obj/item/doctor_bag.dm:115:error: L.insurance: undefined var

BurgerLUA:
Fixed in latest now.

PR #695, opened 2024-08-24
Makes it so examining your skills/attributes with more survival levels gives you tips about them

Gboster-0:
# What this PR does

Makes it so skills/attributes by default have very vague descriptions, with your survival expirience increasing the detail in them.
Also adds in a framework for adding tips to skills/attributes, also locked behind survival expirience of course

The survival skill required to unlock the description of a skill/attribute was chosen by taking how early you level it up, and only 2 of the same number exist in the same file.

Currently the tips are:
SKILL_MELEE
- level 15: "You can perform a strong swing by holding alt and clicking on an enemy.\n
This deals 2 times the damage but requires a wind-up and consumes some stamina."

SKILL_BLOCK
- level 7: "When blocking, if an enemy attacks you in 1 second you will parry them, avoiding all damage and possibly stunning the opponent."
- level 8: "Perfect parrying whilst avoiding all damage, does not grant you blocking expirience."

SKILL_SUMMONING
- level 30: "Your summoning slots are expanded as you level up summoning, currently you can maintain [math] creatures"

Currently survival does not do what it states on the tin, and i think it would be neat adding in tips to skills that tell you about obscure things

### DIFF
--- a/code/_core/datum/experience/_experience.dm
+++ b/code/_core/datum/experience/_experience.dm
@@ -40,7 +40,7 @@
 	var/current_tip = 1
 	for(var/i in 1 to length(tips))
 		if(examine_level >= examine_list[current_tip + 1])
-			completed_list[4] += span("notice", "\n[tips[current_tip]]")
+			completed_list[4] += span("notice", "\n- [tips[current_tip]]")
 		current_tip++
 
 	if(examine_level < examine_list[length(examine_list)])
--- a/code/_core/datum/experience/_experience.dm
+++ b/code/_core/datum/experience/_experience.dm
@@ -5,6 +5,12 @@
 	var/desc = "This is the base experience tracker. You shouldn't have this."
 	var/desc_extended = "Extended description for that wall of text."
 
+	/// If your survival skill is lower than this value, you will see a vague description instead of a normal extended description
+	var/skill_for_examine = 0
+	var/desc_vague = "Affects... something"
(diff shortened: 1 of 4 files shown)


BurgerLUA:
I'm a little conflicted with making the survival skill give information that's already available. I intended it to be used for stuff like examining the current health value of mobs, and not something that anyone can just look up on the repo. The updated descriptions are nice though.

Gboster-0:
> I'm a little conflicted with making the survival skill give information that's already available. I intended it to be used for stuff like examining the current health value of mobs, and not something that anyone can just look up on the repo. The updated descriptions are nice though.

Thats my plan for it for another PR, but imo some of these would be very usefull to have in-game and something that would be annoying to search up, or some more minor tips as most people will never touch the github code.

What i did with blocking and summoning is give the actual end result of calculations of how many minions you have, or how much your blocking is multiplied by via your skills. This is annoying to calculate.

I basically also seperated skill examines into 3 categories that are not really relevant to the PR itself
- "examines" are just the examinations you used to have, a fairly accurate telling of what the skill does
- "Tips" are also what they seem, tips about how you can use certain mechanics (like alt+clicking with a melee weapon) to deal a critical strike on an opponent. They are not really explained in-game (you will forget you can crit strike since its ONLY said in the tutorial book)
- "meta-info" is the most important, and gated behind a very high survival skill, here's where the summoning and defence values are shown to the player

BurgerLUA (approved):

PR #696, opened 2024-08-25
Fixes ore's icon and value

me8my:
# What this PR does
Fixes the icon of ores, and fixes the value of stacks of ores

Makes ores much easier to use, and worth more.

### DIFF
--- a/code/_core/obj/item/material/_material.dm
+++ b/code/_core/obj/item/material/_material.dm
@@ -50,6 +50,7 @@
 /obj/item/material/get_base_value()
 	var/material/M = SSmaterials.all_materials[material_id]
 	. = M.value_per_unit * material_multiplier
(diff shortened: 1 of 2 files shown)


BurgerLUA (approved):

code/_core/obj/item/material/_material.dm
```diff
@@ -50,6 +50,7 @@
 /obj/item/material/get_base_value()
 	var/material/M = SSmaterials.all_materials[material_id]
 	. = M.value_per_unit * material_multiplier
+	. *= amount
```
BurgerLUA, line 53: Actually I'm 90% sure that the amount is already considered in final value calculations.

BurgerLUA (changes requested):

PR #698, opened 2024-08-27
Adds special endings for the crew completing none, or some of the objectives. Syndicate victories are now player-obtainable

Gboster-0:
# What this PR does

- Makes it so world.end() automatically calculates what ending it should have, if not given one

Adds in the following endings:
- Boring Victory
Condition: No objectives were failed nor completed. Total nothingness

- Minor Nanotrasen Victory
Condition: some of the objectives are completed

Changes the following endings conditions to appear:
- Nanotrasen Victory
Condition: All objectives are completed

- Syndicate Victory
Condition: if a lot of objectives were failed and not enough completed
example of when it happens:
0 completed, 1 failed
1 completed, 3 failed
2 completed, 5 failed

Imo its a bit boring that EVERY victory is a nanotrasen victory without admin intervention.
Thus i think making it so players have an impact on the ending is fun.


Stalkeros2:
I *would* approve this PR, but I am uncertain if something like this'd follow Burger's ideas.

BurgerLUA (approved):
I approve, I'm just worried there might be a bug somewhere but lets find out.

PR #699, opened 2024-08-28
Makes creatures that exit a soulgem inherit its soul power, stopping infinite enchantment recharge + sprite fixes

Gboster-0:
# What this PR does

- Soulgems when having power sucked out of a soul will now properly set that soulpower to the mob, making it so you can no longer repair your enchantment, pop out a gabriel out of your gem, pop them back in and repeat so endlessly

- Fixes throwing out unbreaking gems not properly updating their sprites to be empty

1'st point feels like a bug or a weird mechanic, assuming its a bug and squashing it
2'nd annoys me personally, and is 100% a bug


BurgerLUA (approved):
sounds like I need to look at soulgems again to see if they're functioning as intended

Stalkeros2 (approved):
I will assume that the above comment was meant to be interpreted as "Yeah this is good but I'll change it later myself" (because that's what it usually means duh).

PR #700, opened 2024-08-29
Makes NPC's seek out players again

Gboster-0:
# What this PR does

Moves one var check back to where it was, specifically for if the AI has an objective to attack currently

This fixes random elimination squads/enemies chilling until you attack them

In the same commit that burger made AI's run away at round end, they tried to make the code a bit neater by pushing back `can_attack` and `objective_attack` 1 proc backwards.
That in fact made every AI unable to find players, ever, as long as they were not attacked.

### DIFF
--- a/code/_core/datum/ai/ai_life.dm
+++ b/code/_core/datum/ai/ai_life.dm
@@ -37,7 +37,7 @@
 		owner.resist()
 		return TRUE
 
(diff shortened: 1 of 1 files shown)


BurgerLUA (approved):
hilarious and explains why during testing, AIs were not attacking me. I thought it was because I was running debug mode and they could not see me because there was no lighting subsystem running

PR #701, opened 2024-08-29
Makes chemical infused bandages much better

Gboster-0:
# What this PR does

- Medicine(bandages) no longer use a scaling formula, they now always inject the highest amount of chemicals (per 1 bandage) possible
- Medicine no longer is hard-coded to set its max reagent amount to item_amount*10 upon treatment, its now correctly set to the max amount a medical item allows, squashing the bug that made treatment kits overflow at a certain point
- Bandages now properly can only hold 10 units of chemical per bandage

1. Its annoying working with an endlessly changing amount of transferred chemicals, hell it even leads to your treatment kits having 2 times the allowed amount of chemicals.
When you infuse them with 150u at 10 amount, and then treat yourself till the last one remains, at that point you will have 30u per that one treatment kit despite them having 15u max

2. Bug squish make squish sounds

3. Hee hee, bug go \*splat\*

### DIFF
--- a/code/_core/obj/item/container/medicine/_medicine.dm
+++ b/code/_core/obj/item/container/medicine/_medicine.dm
@@ -38,15 +38,15 @@
 
 /obj/item/container/healing/Generate(var/desired_loc)
 	. = ..()
-	if(reagents) reagents.volume_max = amount*reagent_max_per_amount
+	if(reagents) reagents.volume_max = amount * reagent_max_per_amount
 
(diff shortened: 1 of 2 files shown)


Gboster-0:
Actually the overflow bug is still here, resolved at next commit

Stalkeros2 (approved):
I'll pretend to be a competent maintainer and approve this (I mean, CI didn't fail duh).

PR #702, opened 2024-08-29
Makes boss pinpointers no longer show dead bosses

Gboster-0:
# What this PR does

The boss pinpointers now check for the health of the boss, if its less than 0 its excluded from da list

Its annoying tuning your pinpointer onto a boss only to find its dead, this should somewhat solve that


Stalkeros2 (approved):
real

PR #706, opened 2024-11-28
Fixes a very old bug with players being examined

Gboster-0:
# What this PR does

- Fixes the very old bug of players having the description "Level X Your name here.", This is due to the starting name of humanoids being "Your name here.". Now it instead displays the species in that place
- Examining a human no longer even tries to try and take their picture, it never worked, they are too "special"

Before
Before

After
After

Its a bit weird to have the "Level 55 your name here" examine text show up

### DIFF
--- a/code/_core/mob/living/advanced/player/_player.dm
+++ b/code/_core/mob/living/advanced/player/_player.dm
@@ -116,9 +116,10 @@
 	expiration_time = SECONDS_TO_DECISECONDS(180)
 
 /mob/living/advanced/player/get_examine_list(var/mob/living/examiner)
+	var/species/owner_species = SPECIES(species)
 	. = list(
 		div("examine_title", src.name),
-		div("center bold","Level [level] [SPECIES(species)]"),
+		div("center bold","Level [level] [owner_species.name]"),
 		div("examine_description_long", src.desc_extended)
 	)
 
--- a/code/_core/mob/living/advanced/player/_player.dm
+++ b/code/_core/mob/living/advanced/player/_player.dm
@@ -116,17 +116,9 @@
 	expiration_time = SECONDS_TO_DECISECONDS(180)
 
 /mob/living/advanced/player/get_examine_list(var/mob/living/examiner)
-	var/list/species_to_names = list(
-		"human" = "human",
-		"reptile_advanced" = prob(10) ? "spine dragger" : "lizard",
-		"cyborg" = prob(10) ? "tin can": "cyborg",
-		"diona" = "dionae",
-		"moth" = prob(10) ? "moff" : "moth"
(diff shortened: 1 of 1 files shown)


code/_core/mob/living/advanced/player/_player.dm
```diff
@@ -115,6 +115,29 @@
 
 	expiration_time = SECONDS_TO_DECISECONDS(180)
 
+/mob/living/advanced/player/get_examine_list(var/mob/living/examiner)
+	var/list/species_to_names = list(
```
BurgerLUA, line 119: You can get the species from its id by using the define SPECIES(x).

If you want to add funny examine names, you should add that proc to the /species/ datum instead of hardcoding it in a list.

BurgerLUA (changes requested):

code/_core/mob/living/advanced/player/_player.dm
Gboster-0: huh, neat. Applied it

code/_core/mob/living/advanced/player/_player.dm
```diff
@@ -116,17 +116,9 @@
...
-		"reptile_advanced" = prob(10) ? "spine dragger" : "lizard",
-		"cyborg" = prob(10) ? "tin can": "cyborg",
-		"diona" = "dionae",
-		"moth" = prob(10) ? "moff" : "moth"
-	)
-
 	. = list(
 		div("examine_title", src.name),
-		div("center bold","Level [level] [species_to_names[species]]"),
+		div("center bold","Level [level] [SPECIES(species)]"),
```
BurgerLUA, line 121: SPECIES(species) gets the /datum/, not the name of it. I should've been clearer so that's my bad.

BurgerLUA (changes requested):

code/_core/mob/living/advanced/player/_player.dm
Gboster-0: when you display datums, it shows names. So not really a big deal, works

Gboster-0: Yeah displaying a reference to something in the chat just shows its name. Want it to be changed to species.name for clarification or leave it as-is? There's no real difference

BurgerLUA (approved):

code/_core/mob/living/advanced/player/_player.dm
BurgerLUA: Should be species.name because datums don't usually have names, but somehow it works like that. A future update could mess it up so it should be species.name.

BurgerLUA (approved):

PR #711, opened 2025-02-03
Adds a config option to skip non-war phases of the mission gamemode, and a config option to skip vote times. So developers dont go insane

Gboster-0:
# What this PR does

Adds 2 new config options in a new category inside of the config file named "Developer options"
First is `ENABLE_QUICKSTART` that skips all of the mission phases except for "WAR" on the mission gamemode
Second is `ENABLE_QUICKVOTE` that makes all votes last 0.1 seconds

I feel like this would be really handy to have when testing stuff on local servers instead of pressing the progress mission button several times, and waiting for the vote since that DOSENT have a way to skip it as-is


BurgerLUA:
Good idea, but for now you can use force-vote-end (I think that's what it's called) to force the current vote to end.

Gboster-0:
oh neat, didnt know we had that.
But eighter way what i did before this PR is just hard-coded all the times to 0.1 seconds and stash it when uploading.

PR #714, opened 2025-02-09
Major doctor bag rebalance

Gboster-0:
# What this PR does

- First off, doctor bags were bugged, so they would not lose quality, ever, even after 1000 uses because their maximum integrity was not set above normal integrity.
This fixes that and makes it so frogging the doctor bag has a chance to reach the mystical above 190 quality to get an infinite doctor bag, just like before.

- Secondly, makes doctor bags lose 1 quality per use. This is a VERY steep increase in quality cost (ignoring that it was infinite before), but its able to be justified with the next thingy.

- NEW HEALING MODE: BROAD. It lets you heal all of the limbs on a player at once, but your healing is 4 times less efficient
- (This is to adress the issue of very good medicine skill being able to heal 1K+ HP on a limb, whilst the player as a whole has 600 health, and EVERY limb on that person is damaged a little bit)

With infinite integrity, bugs bad.

With the 1 quality per use drop, it actually becomes something you have to watch out for and encurages to get gambling for the mystical 190+ integrity medical doctor bag.

Its very annoying to switch to someones 11 limbs all individually, especially with then possibly running around cancelling your do_after bar several times.
Making it heal all limb damage at once is very handy, can get a bigger healing penalty if 4 times is not enough.

### DIFF
--- a/code/_core/obj/item/doctor_bag.dm
+++ b/code/_core/obj/item/doctor_bag.dm
@@ -2,13 +2,14 @@
 	name = "doctor's bag"
 	desc = "HOLD STILL!"
 	desc_extended = "A highly advanced set of surgery tools and general medical equipment to make you excel at your duties of being a doctor. \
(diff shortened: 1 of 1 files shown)


Stalkeros2 (approved):
Sometimes I forget that I'm a maintainer here; it compiled and the numbers look good so I hope it works.

PR #723, opened 2026-04-05
Fixes max tempering

DaworseplayerEver:
# What this PR does
So this just changes a few numbers around in the code because condition increasing items are supposed to increase everything by 5 percent. However max condition only increased by 1, thus causing a conflict where max condition is increased by 1, but normal conditions and everything increases by 5 without adding to the max.

Max condition and general condition increasing doesnt make sense. the normal condition increases by 5 however max increases only by 1. With this math and how objects can be upgraded to 200 percent now. If you were to upgrade an item at 100 percent condition. It would take at LEAST a HUNDRED items of whatever the condition tempering item is. Such as the bronze weapon kit, whetstone, ETC. That is EXTREMELY grindy and unfun for players especially when they want an upgrade. With this it should fix the problem and get everything back to working. With this it could also open the possibly for more advanced upgrading tools.

### DIFF
--- a/code/_core/obj/item/tempering/tempering_quality.dm
+++ b/code/_core/obj/item/tempering/tempering_quality.dm
@@ -73,7 +73,7 @@
 
 	increase = 5
 	minimum = 100
-	increase_maximum = 1
+	increase_maximum = 5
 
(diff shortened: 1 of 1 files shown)


DaworseplayerEver:
I think my dumbass forgor to compile the files first. then edit so... yeah. I think. I think i am dumb.

